PR #7, opened 2023-05-10
Adapt `NetworkPolicy`s

rfranzke:

/area security networking
/kind enhancement
/platform aws

This PR adapts the `NetworkPolicy`s as following:

- The `extension-shoot-lakom-service` `Service` is now annotated with `networking.resources.gardener.cloud/from-all-webhook-targets-allowed-ports` to allow it being reached from the `kube-apiserver` of shoots (see https://github.com/gardener/gardener/pull/7907). The webhook mutating the pod labels of the `kube-apiserver` is deprecated and can be cleaned up in a future release.
- The custom pod label selectors for Prometheus is adapted similar to https://github.com/gardener/gardener/pull/7907/commits/b14706780cb74902404aa44351bd39124aee48ae.

Related to https://github.com/gardener/gardener/issues/7352

**Release note**:

```other operator
NONE
```

### DIFF
--- a/example/controller-registration.yaml
+++ b/example/controller-registration.yaml
@@ -17,7 +17,8 @@ metadata:
 spec:
   deployment:
     deploymentRefs:
-    - name: shoot-lakom-service
+      - name: shoot-lakom-service
+    policy: Always
   resources:
-  - kind: Extension
-    type: shoot-lakom-service
+    - kind: Extension
+      type: shoot-lakom-service


charts/gardener-extension-shoot-lakom-service/templates/service.yaml
```diff
@@ -10,6 +10,10 @@ metadata:
   namespace: {{ .Release.Namespace }}
   annotations:
     networking.resources.gardener.cloud/from-world-to-ports: '[{"protocol":"TCP","port":{{ .Values.controllers.serverPort }}}]'
+    networking.resources.gardener.cloud/from-all-seed-scrape-targets-allowed-ports: '[{"port":{{ .Values.controllers.metricsPort }},"protocol":"TCP"}]'
+    # TODO: This label approach is deprecated and no longer needed in the future. Remove them as soon as gardener/gardener@v1.75 has been released.
```
vpnachev, line 14: Do you mean annotation, not labels?
And as it is deprecated, why is it added now and needs to be removed in several weeks again?

pkg/controller/webhook/mutate_kube_apiserver.go
```diff
@@ -25,6 +25,7 @@ func (m *kubeAPIServerMutator) Default(_ context.Context, obj runtime.Object) er
 		return fmt.Errorf("expected *appsv1.Deployment but got %T", obj)
 	}
 
+	// TODO: This label approach is deprecated and no longer needed in the future. Remove it as soon as gardener/gardener@v1.75 has been released.
```
vpnachev, line 28: As this webhook is only injecting this label, that means the entire webhook can be removed, right?
Which version of gardener adds the support for the alternative network policy configuration? I mean, I don't have to wait gardener to completely remove support for the deprecated labels/annotations in v1.75, I guess I can remove it even earlier.

charts/gardener-extension-shoot-lakom-service/templates/service.yaml
rfranzke: Yes, annotations of course, fixed it.
I think these annotations are already missing right now (when Gardener's `FullNetworkPolicyInRuntimeCluster` feature gate is enabled, `seed-prometheus` cannot talk to this extension without them).

pkg/controller/webhook/mutate_kube_apiserver.go
rfranzke: > As this webhook is only injecting this label, that means the entire webhook can be removed, right?

Yep, clarified it in the TODO statement.

> Which version of gardener adds the support for the alternative network policy configuration?

Probably v1.71 (depends on https://github.com/gardener/gardener/pull/7907).

> I mean, I don't have to wait gardener to completely remove support for the deprecated labels/annotations in v1.75, I guess I can remove it even earlier.

Sure, that's possible. Operators must ensure that at least g/g@v1.71 is deployed when a version of this extension is used that has removed the old deprecated annotations.

vpnachev (changes requested):

charts/gardener-extension-shoot-lakom-service/templates/service.yaml
```diff
@@ -10,6 +10,10 @@ metadata:
   namespace: {{ .Release.Namespace }}
   annotations:
     networking.resources.gardener.cloud/from-world-to-ports: '[{"protocol":"TCP","port":{{ .Values.controllers.serverPort }}}]'
+    networking.resources.gardener.cloud/from-all-seed-scrape-targets-allowed-ports: '[{"port":{{ .Values.controllers.metricsPort }},"protocol":"TCP"}]'
```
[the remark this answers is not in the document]
rfranzke: Thanks for checking and finding this, I indeed forgot to add the labels that allow ingress from garden namespace. 👏🏻

example/controller-registration.yaml
```diff
@@ -17,8 +17,7 @@ metadata:
...
   deployment:
     deploymentRefs:
-      - name: shoot-lakom-service
-    policy: Always
+    - name: shoot-lakom-service
   resources:
-    - kind: Extension
-      type: shoot-lakom-service
+  - kind: Extension
+    type: shoot-lakom-service
```
vpnachev, line 23: Trimming these white spaces seems to be related to the yq used here https://github.com/gardener/gardener-extension-shoot-lakom-service/blob/02c395c88cc5bc8f939ad50c568142c296ccedf9/hack/set-controller-deployment-policy-to-always.sh#L14. 

I just double checked, and the script is using the `yq` installed in the `hack/tools/bin`. The change was interoduced with f16753958fce80fa4bd1f8e89f9cc2a3af2f1b34 this morning, can you check if something went wrong on your machine?

vpnachev (approved):
/lgtm

PR #16, opened 2023-06-02
Clean a healthcheck todo and register healthcheck for the shoot managedresource

vpnachev:
Clean a healthcheck todo and register healthcheck for the shoot managedresource 
Fixes #

**Release note**:

```other operator
This extension is now also doing health check on the managed resource installing the resources in the shoot cluster.  
```

### DIFF
--- a/pkg/controller/healthcheck/add.go
+++ b/pkg/controller/healthcheck/add.go
@@ -45,8 +45,7 @@ func RegisterHealthChecks(mgr manager.Manager, opts healthcheck.DefaultAddArgs) 
 				HealthCheck:   general.CheckManagedResource(constants.ManagedResourceNamesSeed),
 			},
 		},
-		// TODO(shafeeqes): Remove this condition in a future version.
-		sets.New(gardencorev1beta1.ShootSystemComponentsHealthy),
+		sets.New[gardencorev1beta1.ConditionType](),
 	)
 }
 
--- a/pkg/controller/healthcheck/add.go
+++ b/pkg/controller/healthcheck/add.go
@@ -44,6 +44,10 @@ func RegisterHealthChecks(mgr manager.Manager, opts healthcheck.DefaultAddArgs) 
 				ConditionType: string(gardencorev1beta1.ShootControlPlaneHealthy),
 				HealthCheck:   general.CheckManagedResource(constants.ManagedResourceNamesSeed),
 			},
+			{
+				ConditionType: string(gardencorev1beta1.ShootSystemComponentsHealthy),
+				HealthCheck:   general.CheckManagedResource(constants.ManagedResourceNamesShoot),
+			},
 		},
 		sets.New[gardencorev1beta1.ConditionType](),
 	)


dimityrmirchev (approved):
/lgtm

pkg/controller/healthcheck/add.go
```diff
@@ -44,9 +44,12 @@ func RegisterHealthChecks(mgr manager.Manager, opts healthcheck.DefaultAddArgs)
 				ConditionType: string(gardencorev1beta1.ShootControlPlaneHealthy),
 				HealthCheck:   general.CheckManagedResource(constants.ManagedResourceNamesSeed),
 			},
+			{
+				ConditionType: string(gardencorev1beta1.ShootSystemComponentsHealthy),
+				HealthCheck:   general.CheckManagedResource(constants.ManagedResourceNamesShoot),
```
shafeeqes, line 49: Why was this added back? The shoot care controller already health-checks all the ManagedResources in the shoot control-plane which does not specify `spec.class`, so these MRs are already covered.
For more details, see https://github.com/gardener/gardener/pull/7462.

The cleanup code was to remove this condition from the extensions, since the extensions do not need to maintain this condition anymore.
We just had to remove the cleanup code in this release, not add the health check back.

vpnachev: Good question. I am not aware of all changes in g/g and cannot keep myself informed for any single change there. 

From my point of view, when I am registering a healthcheck for the one managed resource, I asked myself why I am not doing the same for the other managed resource too. Looking into the documentation of https://pkg.go.dev/github.com/gardener/gardener@v1.71.3/extensions/pkg/controller/healthcheck#DefaultRegistration and https://pkg.go.dev/github.com/gardener/gardener@v1.71.3/extensions/pkg/controller/healthcheck#ConditionTypeToHealthCheck


I do not see any recommendation that I must not register healthchecks for `class=shoot`. Also, when I tested the change - I have not seen any error or warning in the controller logs that I am registering a healthcheck that I should not. That is something that you should think about how to improve because now you noticed the change, but I am sure in 2 years nobody would have noticed it.

shafeeqes: From the release notes of gardener v1.65,
```
Now by default, Gardener performs health check for all the `ManagedResource`s with `.spec.class=nil` created in the shoot namespaces. Extensions using Gardener `v1.65.0` onwards can drop the health check for the MangedResource.
```
It's also mentioned in https://github.com/gardener/gardener/blob/32baa826773add7c273ac185b5e15e3ef5d536c4/docs/extensions/healthcheck-library.md?plain=1#L99.

Adding the healthcheck to the extension won't break anything, but just that it is not necessary.

vpnachev: After you noticed this yesterday I also found https://github.com/gardener/gardener/blob/32baa826773add7c273ac185b5e15e3ef5d536c4/docs/extensions/healthcheck-library.md?plain=1#L99 and to be honest even if I was aware about it, I would have registered again the health check as it does not provide any guidance, just claims that Gardener does the health check but there is no statement that the extension controller should/must not do the same health check.

PR #20, opened 2023-06-13
Use Background context instead of TODO in tests

vpnachev:
Use Background context instead of TODO in tests

Fixes #

**Release note**:

```other operator
NONE
```


dimityrmirchev (approved):
/lgtm

PR #22, opened 2023-06-28
Remove usage of policy/v1beta1 API

vpnachev:
Remove usage of policy/v1beta1 API

With https://github.com/gardener/gardener/pull/8087 clusters with k8s version <= 1.21 are not supported, hence there is no need different API versions for PDBs to be used.
Fixes #

**Release note**:

```other operator

```


shafeeqes (commented):
Thanks, lgtm.

pkg/controller/lifecycle/actuator.go
```diff
@@ -482,7 +469,17 @@ func getSeedResources(lakomReplicas *int32, namespace, genericKubeconfigName, sh
...
 		lakomDeployment,
-		lakomPDB,
+		&policyv1.PodDisruptionBudget{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      constants.ExtensionServiceName,
+				Namespace: namespace,
+				Labels:    getLabels(),
+			},
+			Spec: policyv1.PodDisruptionBudgetSpec{
+				MaxUnavailable: intstr.ValueOrDefault(nil, intstr.FromInt(1)),
```
dimityrmirchev, line 479: Nit: I would rather not wrap this in another function call. I find it easier and simpler to read

```golang
maxUnavailable = intstr.FromInt(1)
// and then 
MaxUnavailable: &maxUnavailable,
```

or directly 
```golang
 MaxUnavailable: &intstr.IntOrString{Type: intstr.Int, IntVal: 1}
```

dimityrmirchev (commented):
@vpnachev 
Looks good! I have just one nit for which I leave the decision to you if you want to address this or not.

pkg/controller/lifecycle/actuator.go
vpnachev: Done

dimityrmirchev (approved):
/lgtm

PR #24, opened 2023-07-06
Introduce persona-based docs

vlerenc:
We would like to introduce persona-based docs focussing on the already known release notes personas and therefore "cleanup" the structure somewhat:
- users (`/docs/usage`)
- operators (`/docs/operations`)
- developers (`/docs/development`)

cc @n-boshnakov

**Release note**:
None as ~30 PRs have/will be opened and the message would be the same everywhere (no relevant content changes either) while the only thing that actually changes in the end is the docs page that will have its release note.

/kind/cleanup


.docforge/manifest.yaml
```diff
@@ -1,9 +1,21 @@
-# SPDX-FileCopyrightText: 2022 SAP SE or an SAP affiliate company and Gardener contributors
-#
-# SPDX-License-Identifier: Apache-2.0
```
vpnachev, line 3: Can we keep the license header?

vpnachev (commented):
Do you have some documentation/guideline about the different categories? It would be nice to refer to it when a new documentation is written.

.docforge/manifest.yaml
vlerenc: Sure. Stupid question maybe, what's the background/why?

I found it only in 2 places I believe and since these are our files, I wasn't sure what's the benefit.

vlerenc:
> Do you have some documentation/guideline about the different categories? It would be nice to refer to it when a new documentation is written.

Let's say, it's work in progress, so not yet (and there are also open questions). We also want to have some minimal docs and develop it from there.

.docforge/manifest.yaml
vpnachev: Due to some standards ( I guess [reuse](https://reuse.software/)) or maybe licensing, we are required to have the license header on all possible files.

vlerenc: OK, I added the folder to the reuse manifest, so that we can remove these less than helpful headers. `reuse lint` isn't showing any errors anymore. However, if `reuse` compliance is intended, there are some unrelated (to this PR unrelated) issues:
```
* Files with copyright information: 5732 / 7791
* Files with license information: 488 / 7791

Unfortunately, your project is not compliant with version 3.0 of the REUSE Specification :-(
```

vlerenc:
@vpnachev Is it now ok to merge the PR?

vpnachev (approved):
/lgtm

vlerenc:
Thx @vpnachev !

Issue #25, opened 2023-07-17
Admission webhook caches false resuts?


vpnachev:
**What happened**:
Admission request for pod was canceled and for the next several minutes all following admission requests were rejected because there is no valid signature. However, the image is correctly signed and trusted by the configured public key, so maybe failed signature check was cached as image with no trusted signature. 

**What you expected to happen**:
Internal failures to not be cached.

**How to reproduce it (as minimally and precisely as possible)**:
N/A

**Anything else we need to know**:

<details>
  <summary>spoiler for lakom logs:</summary>

```json
{
    "level": "info",
    "ts": "2023-07-17T11:47:22.428Z",
    "logger": "cosign-signature-verifier",
    "msg": "Handling new pod request",
    "pod": "my-namespace/my-app-0"
}
{
    "level": "info",
    "ts": "2023-07-17T11:47:22.965Z",
    "logger": "cosign-signature-verifier",
    "msg": "image signature successfully verified",
    "pod": "my-namespace/my-app-0",
    "containers": "foo-bar-sidecar",
    "image": "registry.example.com/foo-bar-sidecar@sha256:abcd",
    "containerName": "foo-bar-sidecar",
    "cosignKeyID": 0
}
{
    "level": "info",
    "ts": "2023-07-17T11:47:23.390Z",
    "logger": "cosign-signature-verifier",
    "msg": "no matching signatures found for current public key",
    "pod": "my-namespace/my-app-0",
    "containers": "foo-bar",
    "image": "registry.example.com/foo-bar@sha256:123",
    "containerName": "foo-bar",
    "cosignKeyID": 0,
    "error": "no matching signatures:\nGet \"https://registry.example.com/v2/foo-bar/blobs/sha256:123\": context canceled"
}
{
    "level": "info",
    "ts": "2023-07-17T11:47:23.390Z",
    "logger": "cosign-signature-verifier",
    "msg": "image signature verification failed for all configured keys",
    "pod": "my-namespace/my-app-0",
    "containers": "foo-bar",
    "image": "registry.example.com/foo-bar@sha256:123",
    "containerName": "foo-bar"
}
{
    "level": "error",
    "ts": "2023-07-17T11:47:23.390Z",
    "logger": "cosign-signature-verifier",
    "msg": "pod validation failed",
    "pod": "my-namespace/my-app-0",
[...]

PR #28, opened 2023-07-28
Mitigate wrongly returned NoMatchingSignature error when context is canceled

vpnachev:
Mitigate wrongly returned `NoMatchingSignature` error when context is canceled

Mitigates #25 

**Release note**:

```bugfix operator
A bug that was caching image signature verification status as unsigned due to wrongly returned `NoMatchingSignature` by the SDK is now mitigated. 
```

### DIFF
--- a/pkg/lakom/verifysignature/verifier.go
+++ b/pkg/lakom/verifysignature/verifier.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"crypto"
 	"fmt"
+	"strings"
 
 	"github.com/gardener/gardener-extension-shoot-lakom-service/pkg/constants"
 	"github.com/gardener/gardener-extension-shoot-lakom-service/pkg/lakom/metrics"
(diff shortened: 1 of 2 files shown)


pkg/lakom/verifysignature/verifier.go
```diff
@@ -86,6 +87,12 @@ func verify(ctx context.Context, imageRef name.Reference, keys []crypto.PublicKe
 			}
 
 			if IsNoMatchingSignature(err) {
+				if strings.Contains(err.Error(), "context canceled") {
```
dimityrmirchev, line 90: Nit: I think that this is the preferred way to check for the context canceled error, but since this is some mitigation code both work for me.
```suggestion
				if errors.Is(err, context.Canceled) {
```

dimityrmirchev (approved):

dimityrmirchev (approved):
/lgtm

PR #31, opened 2023-08-14
Remove temporary code cleaning health condition from Extension resource

vpnachev:
Remove temporary code cleaning health condition from Extension resource

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/pkg/controller/healthcheck/add.go
+++ b/pkg/controller/healthcheck/add.go
@@ -15,7 +15,6 @@ import (
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


pkg/controller/healthcheck/add.go
```diff
@@ -46,7 +45,7 @@ func RegisterHealthChecks(mgr manager.Manager, opts healthcheck.DefaultAddArgs)
 				HealthCheck:   general.CheckManagedResource(constants.ManagedResourceNamesSeed),
 			},
 		},
-		sets.New(gardencorev1beta1.ShootSystemComponentsHealthy), // TODO(vpnachev), remove this condition removal in a future version.
+		nil,
```
dimityrmirchev, line 48:
```suggestion
		sets.New(),
```

Wouldn't it be better to pass an empty set here so that we guarantee that this argument does not change to something unexpected in the future?

vpnachev: You mean if this argument of `DefaultRegistration` changes to something other than `sets.Set[gardencorev1beta1.ConditionType])` that still accepts nil as parameter?

dimityrmirchev: Yes exactly, basically if `DefaultRegistration` suddenly accepts different argument that can be represented by `nil`

dimityrmirchev (approved):
/lgtm

PR #44, opened 2023-10-17
Keychain ignores missing secrets

### LINKED ISSUE #43: Keychain to ignore non-existing secrets
Sometimes, the pods refer to image pull secrets that simply do not exist, the image tag resolver and signature verifiers can simply ignore those secrets and work only with the available ones. 

The code that needs to be changed is https://github.com/gardener/gardener-extension-shoot-lakom-service/blob/5ccdfc370f1a685563c05e5f224836cb3739a157/pkg/lakom/resolvetag/admission.go#L158-L160 and https://github.com/gardener/gardener-extension-shoot-lakom-service/blob/5ccdfc370f1a685563c05e5f224836cb3739a157/pkg/lakom/verifysignature/admission.go#L179-L181

**Why is this needed**:
To not fail when non-existing image pull secrets are set in the pods spec.

vpnachev:
Keychain ignores missing secrets

Fixes #43 

**Release note**:

```feature user
Lakom admission webhooks now ignore the image pull secrets that does not exist in the cluster. 
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -79,6 +79,7 @@ require (
 	github.com/docker/docker-credential-helpers v0.7.0 // indirect
 	github.com/dsnet/compress v0.0.1 // indirect
 	github.com/emicklei/go-restful/v3 v3.10.1 // indirect
+	github.com/evanphx/json-patch v5.6.0+incompatible // indirect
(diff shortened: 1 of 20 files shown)


dimityrmirchev (approved):
/lgtm

PR #49, opened 2023-11-07
Add option to disable node identity

### LINKED ISSUE #13: Add option to disable node identity based authentication to OCI registry
The keychain used to authenticate the requests to the OCI registry to be configurable to use only the image pull secret and not the  node identity based authentication. 
 
**Why is this needed**:
The currently keychain is also using node identity based authentication via a token available on the node or retrieved from the metadata server. In gardener scenario, the admission controller is running in the Seed while the admitted pods are deployed in the Shoot cluster - both may run in completely different accounts or even infrastructures. For the seed-only scenario the admission controller is running in-cluster, but the trust between the OCI registry account and the account where the seed is deployed might not be established, hence the node based identity authentication may fail. 

So, it would be nice the admission controller to be configured even to not try to use the node based identity authentication when it is supposed always to fail.

vpnachev:
Lakom application now have a new flag `--use-only-image-pull-secrets` that can be used to let it use only the image pull secrets from the pod. 

Lakom extension controller has a configuration option `useOnlyImagePullSecrets` which respectively controls the `--use-only-image-pull-secrets` flag.

Fixes #13 

**Release note**:

```feature user
Lakom application now can be configured via the flag `--use-only-image-pull-secrets` to use only image pull secrets of the pod to authenticate against the OCI registry, i.e. it will not use the node identity or default docker configuration when the flag is set to `true`.
```

```feature operator
Lakom extension controller now has a configuration option `useOnlyImagePullSecrets` which controls the value of the lakom flag `--use-only-image-pull-secrets`.
```


dimityrmirchev (approved):
/lgtm

PR #53, opened 2023-11-15
Keep secretsmanager secrets in the seed during migration

vpnachev:
Keep secretsmanager secrets in the seed during migration

Fixes #

cc @plkokanov 

**Release note**:

```bugfix operator
A bug in the `shoot-lakom-service` controller that was causing the lakom CA secret for a shoot cluster to be recreated instead of restored during control plane migration has been fixed.  
```


dimityrmirchev (approved):
/lgtm

PR #63, opened 2024-01-02
Adapt PDBs to use `unhealthyPodEvictionPolicy: AlwaysAllow`

shafeeqes:

/kind cleanup

- Add `unhealthyPodEvictionPolicy: AlwaysAllow` to PDBs

Part of https://github.com/gardener/gardener/issues/8821

**Release note**:

```breaking operator
NONE
```


pkg/controller/seed/reconciler_test.go
```diff
@@ -71,40 +75,68 @@ hjZVcW2ygAvImCAULGph2fqGkNUszl7ycJH/Dntw4wMLSbstUZomqPuIVQ==
...
-				cosignSecretNameKey:   expectedSecretCosign(namespace, cosignSecretName, cosignPublicKeys),
-				serviceKey:            expectedService(namespace),
-				serviceAccountKey:     expectedServiceAccount(namespace),
-				vpaKey:                expectedVPA(namespace),
-			}
-
-			for key, expectedResource := range expectedResources {
-				resource, ok := resources[key]
-				Expect(ok).To(BeTrue())
+			JustBeforeEach(func() {
```
vpnachev, line 80: Please, rework it to be table test and pass the kubernetes version as parameter. Right now it requires multiple before nodes to be properly synced, while a more simple test with one parameter can be used.

pkg/controller/seed/reconciler_test.go
```diff
@@ -54,6 +55,9 @@ var _ = Describe("Reconciler", func() {
 		var (
 			cosignPublicKeys []string
 			caBundle         = []byte("caBundle")
+			err              error
+			resources        map[string][]byte
```
vpnachev, line 59:
```suggestion
```

With `DesctibeTable` style test won't need these vars available at this scope level.

pkg/controller/seed/reconciler_test.go
```diff
@@ -54,6 +55,9 @@ var _ = Describe("Reconciler", func() {
...
+			k8sVersion       *semver.Version
```
vpnachev, line 60: Set a default value in the `BeforeEach` node below and use it in most of the invocations of `getResources` functions.

pkg/controller/seed/reconciler_test.go
```diff
@@ -417,9 +452,8 @@ status: {}
 `
 }
 
-func expectedPDB(namespace string) string {
-
-	return `apiVersion: policy/v1
+func expectedPDB(namespace string, k8sGreaterEqual126 bool) string {
```
vpnachev, line 455:
```suggestion
func expectedPDB(namespace string, k8sVersion *semver.Version ) string {
```

In case a new special case needs to be handled for e.g. k8s 1.28, it does not look convenient to use another boolean parameter, thus pass the k8s version and parse it in the function.

vpnachev (changes requested):
Thanks! 
I have some change requests for the tests, rest looks good.

vpnachev:
> Drop support for Shoots, Seeds and Gardens with Kubernetes version == 1.24.

Actually, how is this achieved? I see no change that actually is doing it as the extension was not managing any special cases to k8s <= 1.24, also the version of the github.com/gardener/gardener is not updated, hence the [kubernetes client still allows 1.24](https://pkg.go.dev/github.com/gardener/gardener@v1.88.0/pkg/utils/validation/kubernetesversion#pkg-variables)

shafeeqes:
> > Drop support for Shoots, Seeds and Gardens with Kubernetes version == 1.24.
> 
> Actually, how is this achieved? I see no change that actually is doing it as the extension was not managing any special cases to k8s <= 1.24, also the version of the github.com/gardener/gardener is not updated, hence the [kubernetes client still allows 1.24](https://pkg.go.dev/github.com/gardener/gardener@v1.88.0/pkg/utils/validation/kubernetesversion#pkg-variables)

Rather than the g/g version vendored, I think what matters is the version in which the gardener-apiserver is running. 

And yes, you are right about
> as the extension was not managing any special cases to k8s <= 1.24.

I renamed the PR.

pkg/controller/seed/reconciler_test.go
shafeeqes: I chose to do it this way because if we use the version check here as well, then it just does the same thing as the original function which is tested. But I can change it if you insist.

pkg/controller/seed/reconciler_test.go
```diff
@@ -69,43 +71,51 @@ hjZVcW2ygAvImCAULGph2fqGkNUszl7ycJH/Dntw4wMLSbstUZomqPuIVQ==
...
-				serviceKey:            expectedService(namespace),
-				serviceAccountKey:     expectedServiceAccount(namespace),
-				vpaKey:                expectedVPA(namespace),
-			}
-
-			for key, expectedResource := range expectedResources {
-				resource, ok := resources[key]
-				Expect(ok).To(BeTrue())
+		DescribeTable("Should ensure resources are correctly created for different Kubernetes versions",
+			func(k8sVersion string, expectedPDBResult string) {
```
vpnachev, line 78:
```suggestion
			func(k8sVersion string, withUnhealthyPodEvictionPolicy bool) {
```

pkg/controller/seed/reconciler_test.go
```diff
@@ -69,43 +71,51 @@ hjZVcW2ygAvImCAULGph2fqGkNUszl7ycJH/Dntw4wMLSbstUZomqPuIVQ==
...
-		})
+				Expect(err).ToNot(HaveOccurred())
+				Expect(resources).To(HaveLen(10))
+
+				expectedResources := map[string]string{
+					validatingWebhookKey:  expectedValidatingWebhook(caBundle, failurePolicy),
+					mutatingWebhookKey:    expectedMutatingWebhook(caBundle, failurePolicy),
+					clusterRoleKey:        expectedClusterRole(),
+					clusterRoleBindingKey: expectedClusterRoleBinding(),
+					deploymentKey:         expectedDeployment(namespace, image, cosignSecretName, serverTLSSecretName),
```
vpnachev, line 97:
```suggestion
					deploymentKey:         expectedDeployment(namespace, image, cosignSecretName, serverTLSSecretName),
										pdbKey:                expectedPDB(namespace, withUnhealthyPodEvictionPolicy),

```

pkg/controller/seed/reconciler_test.go
```diff
@@ -69,43 +71,51 @@ hjZVcW2ygAvImCAULGph2fqGkNUszl7ycJH/Dntw4wMLSbstUZomqPuIVQ==
...
+				for key, expectedResource := range expectedResources {
+					resource, ok := resources[key]
+					Expect(ok).To(BeTrue())
+
+					strResource := string(resource)
+					Expect(strResource).To(Equal(expectedResource), key, string(resource))
+
+					pdb, ok := resources[pdbKey]
+					Expect(ok).To(BeTrue())
+					Expect(string(pdb)).To(Equal(expectedPDBResult), pdbKey)
```
vpnachev, line 113:
```suggestion
```

PS. Checking the PDB in the for loop was unnecessary 😉

pkg/controller/seed/reconciler_test.go
```diff
@@ -69,43 +71,51 @@ hjZVcW2ygAvImCAULGph2fqGkNUszl7ycJH/Dntw4wMLSbstUZomqPuIVQ==
...
+				}
+			},
+			Entry("Kubernetes version < 1.26", "1.25.0", expectedPDB(namespace, false)),
+			Entry("Kubernetes version >= 1.26", "1.26.0", expectedPDB(namespace, true)),
```
vpnachev, line 117:
```suggestion
			Entry("Kubernetes version < 1.26", "1.25.0", false),
			Entry("Kubernetes version >= 1.26", "1.26.0", true),
```

vpnachev (changes requested):
Please, apply the same test changes to `pkg/controller/lifecycle/actuator_test.go`

pkg/controller/seed/reconciler_test.go
vpnachev: Ok, soudns good.

vpnachev (approved):
/lgtm

PR #74, opened 2024-02-21
Lakom uses warning instead of failure policy `Ignore`

vpnachev:
Lakom uses warning instead of failure policy `Ignore`

Fixes #

**Release note**:

```feature user
Lakom admission controller is extended with a new flag `--insecure-allow-untrusted-images`. When it is set, the admission webhook returns just warning but still allows the images that are not signed or are not signed with trusted keys. 
```

```feature operator
Lakom gardener extension controller configuration has new field `allowUntrustedImages`, it is used to control the lakom admission controller flag `--insecure-allow-untrusted-images`.
```

```breaking operator
:warning: Lakom admission webhooks now always use failure policy `Fail` and it is no longer possible to change it to `Ignore`. If you want to allow untrusted images
- for the extension controller you can set the field `allowUntrustedImages` to true
- for the lakom application you can set the flag  `--insecure-allow-untrusted-images` to true
Both configs are also exposed via the helm charts values and ControllerDeployment config.
```


cmd/lakom/app/app.go
```diff
@@ -99,6 +98,9 @@ type Options struct {
 	// UseOnlyImagePullSecrets sets only the image pull secrets of the pod to be used to access the OCI registry.
 	// Otherwise, also the node identity and docker config file are used.
 	UseOnlyImagePullSecrets bool
+	// SkipSignatureVerification configures the webhook to not disallow images without trusted signature.
+	// Instead to deny the request, the webhook will allow it with a warning.
+	SkipSignatureVerification bool
```
dimityrmirchev, line 103:
```suggestion
	// AllowUnverifiedImages configures the webhook to allow images without trusted signature.
	// Such images are not rejected, but instead allowed with a warning.
	AllowUnverifiedImages bool
```

dimityrmirchev (commented):
Not a strong opinion, so I am fine both ways. 

As a general comment I think that "skip" is not the proper term to use here since we still do the validation, but allow the image even if validation fails. So IMO "allow" should be the better term.

A second note. What is the use case for this change and why it is better than the ability to ignore denied requests as it is now?

cmd/lakom/app/app.go
vpnachev: Good suggestion, `unverified` is also not the best word, but I will think about it.

pkg/controller/lifecycle/actuator_test.go
```diff
@@ -532,19 +537,13 @@ spec:
...
       app.kubernetes.io/name: lakom
-      app.kubernetes.io/part-of: shoot-lakom-service
-`
-	if withUnhealthyPodEvictionPolicy {
-		out += `  unhealthyPodEvictionPolicy: AlwaysAllow
-`
-	}
-	out += `status:
+      app.kubernetes.io/part-of: shoot-lakom-service` + unhealthyPodEvictionPolicyStr + `
+status:
```
dimityrmirchev, line 541: I think that the nicer formatting will be to set the newline here and not in the variable.

```suggestion
      app.kubernetes.io/part-of: shoot-lakom-service
  ` + unhealthyPodEvictionPolicyStr + `
status:
```

pkg/controller/lifecycle/actuator_test.go
```diff
@@ -246,10 +246,9 @@ webhooks:
 `
 }
 
-func expectedSeedValidatingWebhook(caBundle []byte, namespace string, failurePolicy admissionregistrationv1.FailurePolicyType) string {
+func expectedSeedValidatingWebhook(caBundle []byte, namespace string) string {
 	var (
-		caBundleEncoded  = b64.StdEncoding.EncodeToString(caBundle)
-		strFailurePolicy = string(failurePolicy)
+		caBundleEncoded = b64.StdEncoding.EncodeToString(caBundle)
```
dimityrmirchev, line 251:
```suggestion
		caBundleEncoded := b64.StdEncoding.EncodeToString(caBundle)
```

dimityrmirchev (dismissed):
Two small suggestions that are completely optional

/lgtm

pkg/controller/lifecycle/actuator_test.go
vpnachev: This assignment operator is not allowed inside `var ()` block 😉 

```sh
$ make test
...
# github.com/gardener/gardener-extension-shoot-lakom-service/pkg/controller/lifecycle
pkg/controller/lifecycle/actuator_test.go:251:19: expected type, found ':='
```

pkg/controller/lifecycle/actuator_test.go
vpnachev: When `withUnhealthyPodEvictionPolicy == false`, there will be one unneeded empty line just before status and the test will be failing.

The alternative is

```go
func expectedSeedPDB(namespace string, withUnhealthyPodEvictionPolicy bool) string {
	unhealthyPodEvictionPolicyStr := ""
	if withUnhealthyPodEvictionPolicy {
		unhealthyPodEvictionPolicyStr = `  unhealthyPodEvictionPolicy: AlwaysAllow
`
	}
	return `apiVersion: policy/v1
kind: PodDisruptionBudget
metadata:
  creationTimestamp: null
  labels:
    app.kubernetes.io/name: lakom
    app.kubernetes.io/part-of: shoot-lakom-service
  name: extension-shoot-lakom-service
  namespace: ` + namespace + `
spec:
  maxUnavailable: 1
  selector:
    matchLabels:
      app.kubernetes.io/name: lakom
      app.kubernetes.io/part-of: shoot-lakom-service
` + unhealthyPodEvictionPolicyStr + `status:
  currentHealthy: 0
  desiredHealthy: 0
  disruptionsAllowed: 0
  expectedPods: 0
`
}
```

But now `status` is at the end of a line and `unhealthyPodEvictionPolicyStr` have a trailing new line. 
Both options have some prons and cons with regards to formatting, none of them looks cool 😎

pkg/controller/lifecycle/actuator_test.go
dimityrmirchev: I meant removing the `var` keyword since it is a single variable in there and there is no need for grouping

pkg/controller/lifecycle/actuator_test.go
dimityrmirchev: Ah, I see. Both ways are not pleasant to the eye.

pkg/controller/lifecycle/actuator_test.go
vpnachev: Done with e1c2bcb7

dimityrmirchev (approved):
/lgtm

PR #83, opened 2024-04-18
Always scale up during deletion of the shoot

vpnachev:
Always scale up during deletion of the shoot

During deletion of a hibernated cluster, gardener ensures some of the control plane components are up and running in order to delete the in-cluster resources. However, lakom is not scaled up and create/update operations for clusters pods may fail until the extension is deleted.

Fixes #

Even with this change, gardener just deletes the extension without reconciling it, therefore the lakom deployment is not scaled up at all. 

**Release note**:

```other operator

```

### DIFF
--- a/pkg/controller/lifecycle/actuator.go
+++ b/pkg/controller/lifecycle/actuator.go
@@ -75,9 +75,13 @@ type actuator struct {
 	serviceConfig config.Configuration
 }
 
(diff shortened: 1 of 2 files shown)


vpnachev:
/close

As the extensions are not reconciled in the deletion flow, hence this change will be never active.

PR #84, opened 2024-04-18
Change extension controller pod anti affinity to preferred

vpnachev:
Change extension controller pod anti affinity to preferred

This allows the controller to run with multiple replicas on the same node, which is useful for seed clusters having few nodes.

Fixes #

**Release note**:

```other operator

```


charts/gardener-extension-shoot-lakom-service/templates/deployment.yaml
```diff
@@ -45,18 +45,20 @@ spec:
...
-            topologyKey: "kubernetes.io/hostname"
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 100
+            podAffinityTerm:
+              labelSelector:
+                matchExpressions:
+                - key: app.kubernetes.io/name
+                  operator: In
+                  values:
+                  - {{ .Values.name }}
```
dimityrmirchev, line 56:
```suggestion
                  - gardener-extension-{{ .Values.serviceName }}
```

Isn't this the correct value? I do not see {{ .Values.name }} being set as a label on the pods.

dimityrmirchev (changes requested):

charts/gardener-extension-shoot-lakom-service/templates/deployment.yaml
vpnachev: Good catch!

dimityrmirchev (approved):
/lgtm

PR #91, opened 2024-06-05
Verify all images for managedseed

rrhubenov:
When a shoot is a managed seed cluster, all of the resources in the kube-system namespace need to be validated and not only those that are labeled with `managed-by: gardener`.

With this change, if the shoot is a managed seed (determined by checking if it is in the garden namespace) the selector for `managed-by: gardener` won't be used.

Fixes #

Note that this method for handling the case where the cluster is a managed seed will most likely be changed/updated in the near future. 

**Release note**:

```feature operator
All pods in Managed Seed clusters will now be validated for trusted image signatures.
```

### DIFF
--- a/pkg/controller/lifecycle/actuator.go
+++ b/pkg/controller/lifecycle/actuator.go
@@ -104,6 +104,7 @@ func (a *actuator) Reconcile(ctx context.Context, logger logr.Logger, ex *extens
 		return err
 	}
 
+
 	// initialize SecretsManager based on Cluster object
 	configs := secrets.ConfigsFor(namespace)
 
@@ -140,6 +141,7 @@ func (a *actuator) Reconcile(ctx context.Context, logger logr.Logger, ex *extens
 		image.Tag = ptr.To[string](version.Get().GitVersion)
 	}
 
+
 	seedResources, err := getSeedResources(
 		getLakomReplicas(controller.IsHibernationEnabled(cluster)),
 		namespace,
@@ -162,6 +164,7 @@ func (a *actuator) Reconcile(ctx context.Context, logger logr.Logger, ex *extens
 		caBundleSecret.Data[secretutils.DataKeyCertificateBundle],
 		namespace,
 		lakomShootAccessSecret.ServiceAccountName,
+		cluster.Shoot.GetNamespace(),
 	)
 
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/controller/lifecycle/actuator.go
```diff
@@ -629,6 +633,11 @@ func getShootResources(webhookCaBundle []byte, namespace, shootAccessServiceAcco
 		}}
 	)
 
+	isManagedSeed := shootNamespace == v1beta1constants.GardenNamespace
+	if isManagedSeed {
+		objectSelector = nil
```
vpnachev, line 638:
```suggestion
		objectSelector = metav1.LabelSelector{}
```

> An empty label selector matches all objects. A null label selector matches no objects.

ref: https://github.com/kubernetes/kubernetes/blob/9e2075b3c87061d25759b0ad112266f03601afd8/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L1239-L1240

rrhubenov: Thanks!

pkg/controller/lifecycle/actuator.go
```diff
@@ -592,13 +593,13 @@ func getSeedResources(lakomReplicas *int32, namespace, genericKubeconfigName, sh
 	return resources, nil
 }
 
-func getShootResources(webhookCaBundle []byte, namespace, shootAccessServiceAccountName string) (map[string][]byte, error) {
+func getShootResources(webhookCaBundle []byte, extensionNamespace, shootAccessServiceAccountName, shootNamespace string) (map[string][]byte, error) {
```
vpnachev, line 596: The namespace `shoot--<project>--<shoot>` is called shoot or seed namespace, especially in the context of the seed cluster (where usually the extensions are working). To avoid some confusion, let's rename the variable `shootNamespace` to `projectNamespace`.

```suggestion
func getShootResources(webhookCaBundle []byte, extensionNamespace, shootAccessServiceAccountName, projectNamespace string) (map[string][]byte, error) {
```

pkg/controller/lifecycle/actuator_test.go
```diff
@@ -301,6 +320,94 @@ webhooks:
 `
 }
 
+func expectedManagedSeedMutatingWebhook(caBundle []byte, namespace string) string {
```
vpnachev, line 323: Can you enhance the existing `expectedSeed*Webhook` functions instead of implementing new ones? The existing funcs can accept additional boolean parameter whether object selector should be empty or not. 

It could look like 

```go
func expectedShootMutatingWebhook(caBundle []byte, namespace string, withEmptyObjectSelector bool) string {
	caBundleEncoded := b64.StdEncoding.EncodeToString(caBundle)
	objectSelector := ` {}`
	if !withEmptyObjectSelector {
		objectSelector = `
    matchExpressions:
    - key: resources.gardener.cloud/managed-by
      operator: In
      values:
      - gardener`
	}

    return `apiVersion: admissionregistration.k8s.io/v1
...
  objectSelector:` + objectSelector + `
    matchExpressions:
..
}
```

pkg/controller/lifecycle/actuator.go
rrhubenov: Ah, thanks! I guess I'm confusing:
- shoot namespace (meaning the namespace in which resources for this specific shoot are contained)
- project namespace (namespace in which the "shoot" resource is contained)

vpnachev (approved):
/lgtm

PR #93, opened 2024-06-17
Skaffold for local deployment

rrhubenov:
Local deployment using skaffold enables easy local development without a remote cluster.
Fixes #

**Release note**:

```feature developer
Easy local development using skaffold deployments.
```


hack/check-skaffold-deps.sh
```diff
@@ -0,0 +1,28 @@
...
+
+operation="${1:-check}"
+
+echo "> ${operation^} Skaffold Dependencies"
+
+success=true
+repo_root="$(git rev-parse --show-toplevel)"
+
+function run() {
+  if ! "$repo_root"/hack/check-skaffold-deps-for-binary.sh "$operation" --skaffold-file "$1" --binary "$2" --skaffold-config "$3"; then
```
vpnachev, line 17: Isn't it possible to re-use the original script from Gardener similarly to https://github.com/gardener/gardener-discovery-server/blob/main/hack/check-skaffold-deps.sh ?

vpnachev (changes requested):
Thanks!
Looks very nice, I have some minor suggestions.

oliver-goetz (commented):
I like your version with the script and the `after-build` hook.
It is easier to understand than the custom build script we use for machine-controller-manager-provider-local 👍

hack/check-skaffold-deps.sh
rrhubenov: Yeah, nice idea, thanks : )

hack/tools.mk
```diff
@@ -22,3 +24,6 @@ tool_version_file = $(TOOLS_BIN_DIR)/.version_$(subst $(TOOLS_BIN_DIR)/,,$(1))_$
 
 $(COSIGN): $(call tool_version_file,$(COSIGN),$(COSIGN_VERSION))
 	GOBIN=$(abspath $(TOOLS_BIN_DIR)) go install github.com/sigstore/cosign/v2/cmd/cosign@$(COSIGN_VERSION)
+$(CRANE): $(call tool_version_file,$(CRANE),$(CRANE_VERSION))
```
vpnachev, line 27: Just add one empty line to improve readability.

```suggestion

$(CRANE): $(call tool_version_file,$(CRANE),$(CRANE_VERSION))
```

skaffold.yaml
```diff
@@ -0,0 +1,83 @@
...
+  artifacts:
+    - image: local-skaffold/gardener-extension-shoot-lakom-service
+      # The extension controller uses the vectorImageOverwrite values to use 
+      # as the image for the resources it deploys.
+      # This requires that the admission controller is built before this image.
+      requires:
+        - image: local-skaffold/lakom
+      ko:
+        # The dependencies section configures what files Skaffold should watch for changes when in dev mode.
+        # These dependencies can be updated using a skaffold script found in the hack dir of gardener
```
vpnachev, line 19:
```suggestion
        # These dependencies can be updated using a skaffold script found in the hack dir
```

skaffold.yaml
```diff
@@ -0,0 +1,83 @@
...
+            - pkg/controller/seed
+            - pkg/imagevector
+            - pkg/secrets
+            - VERSION
+        ldflags:
+          - '{{.LD_FLAGS}}'
+        main: ./cmd/gardener-extension-shoot-lakom-service
+    - image: local-skaffold/lakom
+      # We need the image for the lakom admisison controller to be injected 
+      # into the vectorImageOverwrite for the controller registration.
```
vpnachev, line 41:
```suggestion
      # into the imageVectorOverwrite for the controller registration.
```

Makefile
```diff
@@ -131,3 +132,24 @@ verify: check format test
 
 .PHONY: verify-extended
 verify-extended: check-generate check format test test-cov test-clean
+
+.PHONY: update-skaffold-deps
+update-skaffold-deps: $(YQ)
+	@GARDENER_HACK_DIR=$(GARDENER_HACK_DIR) hack/check-skaffold-deps.sh update
```
vpnachev, line 138: For consistency, let's use the full path to the hack scripts.

```suggestion
	@GARDENER_HACK_DIR=$(GARDENER_HACK_DIR) $(HACK_DIR)/hack/check-skaffold-deps.sh update
```

Makefile
```diff
@@ -131,3 +132,24 @@ verify: check format test
...
+
+# speed-up skaffold deployments by building all images concurrently
+export SKAFFOLD_BUILD_CONCURRENCY = 0
+extension-up extension-dev: export SKAFFOLD_DEFAULT_REPO = localhost:5001
+extension-up extension-dev: export SKAFFOLD_PUSH = true
+# use static label for skaffold to prevent rolling all gardener components on every `skaffold` invocation
+extension-up extension-dev extension-down: export SKAFFOLD_LABEL = skaffold.dev/run-id=extension-local
+
+extension-up: $(SKAFFOLD) $(KIND) $(HELM) $(KUBECTL)
```
vpnachev, line 147: I think this and `extension-dev` targets should depend on `$(CRANE)`
```suggestion
extension-up: $(SKAFFOLD) $(KIND) $(HELM) $(KUBECTL) $(CRANE)
```

hack/skaffold-imagevector.sh
```diff
@@ -0,0 +1,35 @@
...
+# SKAFFOLD_IMAGE_REPO=localhost:5001 (instead of localhost:5001/lakom)
+# SKAFFOLD_IMAGE_TAG=tag
+#
+# This might need to be brought up as an issue on skaffold's side. For anyone
+# interested, the logic for the parsing of the image name in the skaffold project 
+# is contained in https://github.com/GoogleContainerTools/skaffold/blob/main/pkg/skaffold/docker/reference.go
+#
+# Additionally, we want to inject the image digest from the manifest in the registry.
+# Skaffold does not export this value. That's why crane is used to append it manually.
+IMAGE_REPO=$(echo $SKAFFOLD_IMAGE | cut -d':' -f1,2)
```
vpnachev, line 20: Let's set the `errexit` and `pipefail` options to shell to let it abort the process if any of the commands fail 

```suggestion

set -o pipefail
set -o errexit

IMAGE_REPO=$(echo $SKAFFOLD_IMAGE | cut -d':' -f1,2)
```

hack/skaffold-imagevector.sh
```diff
@@ -0,0 +1,35 @@
...
+IMAGE_TAG=$(echo "$(echo $SKAFFOLD_IMAGE | cut -d':' -f3)@$(crane digest $SKAFFOLD_IMAGE)")
+
+cat <<EOF > example/lakom/skaffold/patch-imagevector-overwrite.yaml
```
vpnachev, line 23:
```suggestion
mkdir -p example/lakom/skaffold
cat <<EOF > example/lakom/skaffold/patch-imagevector-overwrite.yaml
```

vpnachev (changes requested):
Thanks!
Works very nice, I just had to made a few changes locally to run it successfully and also propose them together with few minor nits.

Makefile
```diff
@@ -53,9 +53,10 @@ start-lakom:
 		--cosign-public-key-path=example/lakom/cosign/cosign.pub \
 		--cache-ttl=$(CACHE_TTL) \
 		--cache-refresh-interval=$(CACHE_REFRESH_INTERVAL)
+		--insecure-allow-untrusted-images=true
 
 .PHONE: dev-setup
-dev-setup: $(COSIGN)
+dev-setup: $(COSIGN) $(CRANE)
```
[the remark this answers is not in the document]
rrhubenov: Whoops, meant to add it to `extension-up`. Thanks!

hack/update-skaffold-deps.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
```
vpnachev, line 1: I think this script can be removed in favor of the Makefile target `update-skaffold-deps`.

Makefile
```diff
@@ -131,3 +132,24 @@ verify: check format test
 
 .PHONY: verify-extended
 verify-extended: check-generate check format test test-cov test-clean
+
+.PHONY: update-skaffold-deps
+update-skaffold-deps: $(YQ)
+	@GARDENER_HACK_DIR=$(GARDENER_HACK_DIR) $(HACK_DIR)/check-skaffold-deps.sh update
```
vpnachev, line 138: Call `@GARDENER_HACK_DIR=$(GARDENER_HACK_DIR) $(HACK_DIR)/check-skaffold-deps.sh` from the `check` target.

hack/check-skaffold-deps.sh
```diff
@@ -0,0 +1,28 @@
...
+
+function run() {
+  if ! bash "$GARDENER_HACK_DIR"/check-skaffold-deps-for-binary.sh "$operation" --skaffold-file "$1" --binary "$2" --skaffold-config "$3"; then
+    success=false
+  fi
+}
+
+# skaffold.yaml
+run "skaffold.yaml" "gardener-extension-shoot-lakom-service"      "extension"
+run "skaffold.yaml" "lakom"                                       "admission"
```
vpnachev, line 24:
```suggestion
run "skaffold.yaml" "lakom"                                       "extension"
```

vpnachev (changes requested):
Some minor nits. 

You can also change the year in the license headers in the new files to 2024.

rrhubenov:
I'll be working on  #96 in this PR as well.
Since the local registry does not support HTTPS, we'll need to add a flag to Lakom to allow HTTP communication with the registries. Might be useful for production as well, but it's mainly targeted for local dev.

cmd/lakom/app/app.go
```diff
@@ -211,6 +215,7 @@ func (o *Options) Run(ctx context.Context) error {
 		WithCacheTTL(o.CacheTTL).
 		WithCacheRefreshInterval(o.CacheRefreshInterval).
 		WithUseOnlyImagePullSecrets(o.UseOnlyImagePullSecrets).
+                WithAllowInsecureRegistries(o.AllowInsecureRegistries).
```
vpnachev, line 218: `cosignSignatureVerifyHandler` must implement the same feature because it is also communicating with the same OCI registry.

pkg/lakom/resolvetag/admission.go
```diff
@@ -193,7 +202,15 @@ func (h *handler) handlePod(ctx context.Context, p *corev1.Pod, logger logr.Logg
...
-	imageRef, err := name.ParseReference(image)
+        var (
+            imageRef name.Reference
+            err error
+        )
+        if h.allowInsecureRegistries {
+            imageRef, err = name.ParseReference(image, name.Insecure)
+        } else {
+            imageRef, err = name.ParseReference(image)
+        }
```
vpnachev, line 213: I would propose something like 
```go
	opts := []name.Option{}
	if h.allowInsecureRegistries {
		opts = append(opts, name.Insecure)
	}
	imageRef, err := name.ParseReference(image, opts...)
```

This way there is no else statement and if more options needs to be enabled, it will be straight forward.

vpnachev (changes requested):
Thanks!
Please, add the insecure feature to the image signature verifier as it is also connecting to the OCI registry.

pkg/lakom/resolvetag/admission.go
rrhubenov: Thanks that's much better

pkg/lakom/verifysignature/verifier.go
```diff
@@ -134,7 +138,7 @@ func NewCacheVerifier(cache SignatureVerificationResultCache, verifier Verifier)
 // Verify check cosign signature of an image. Firstly it checks if the cache have an entry
 // for the verification state of the image and returns it. If the cache have no entry,
 // it uses the verifier to do the real verification, persists the result in the cache and return it.
-func (r *cacheVerifier) Verify(ctx context.Context, image string, kcr utils.KeyChainReader) (bool, error) {
+func (r *cacheVerifier) Verify(ctx context.Context, image string, kcr utils.KeyChainReader, insecure bool) (bool, error) {
```
vpnachev, line 141: In theory, the cache verifier should not be aware of the insecure option. It is also not aware of the implementation of the `actualVerifier`, i.e. it could be anything that implements the `Verifier` interface and it might not consider the insecure option. 

I would propose to not add `insecure bool` parameter to the `Verify` method, instead put this feature only in the `directVerifier` and configure it during initialization, just like the cosign public keys are configured. Then only the directVerifier implementation of `Verify` will be aware of the `insecure` option. 

I mean something like:

```go
type directVerifier struct {
	publicKeys []crypto.PublicKey
	insecure   bool
}

// NewDirectVerifier creates new verifier and returns it.
func NewDirectVerifier(keys []crypto.PublicKey, insecure bool) *directVerifier {
	dv := &directVerifier{
		publicKeys: keys,
		insecure:   insecure,
	}
	return dv
}

// Verify check if image is signed by at least one of the configured cosign public keys.
func (r *directVerifier) Verify(ctx context.Context, image string, kcr utils.KeyChainReader) (bool, error) {
	opts := []name.Option{}
	if r.insecure {
		opts = append(opts, name.Insecure)
	}
...
```

vpnachev (commented):
Just one proposal to not change the Verifier interface. 
Also, please change all spaces to tabs in the go files.

pkg/lakom/verifysignature/verifier.go
rrhubenov: I agree, your solution is better. I'll apply it

Makefile
```diff
@@ -53,6 +53,7 @@ start-lakom:
 		--cosign-public-key-path=example/lakom/cosign/cosign.pub \
 		--cache-ttl=$(CACHE_TTL) \
 		--cache-refresh-interval=$(CACHE_REFRESH_INTERVAL)
+		--insecure-allow-untrusted-images=true
```
vpnachev, line 56:
```suggestion
		--cache-refresh-interval=$(CACHE_REFRESH_INTERVAL)
```

or 


```suggestion
		--cache-refresh-interval=$(CACHE_REFRESH_INTERVAL) \
		--insecure-allow-untrusted-images=true
```

vpnachev (changes requested):
Just one last minor change request in the Makefile, rest looks good.

PR #98, opened 2024-07-25
Publish Helm charts as OCI artifacts

oliver-goetz:
/area delivery
/kind enhancement

We should start publishing Helm charts as OCI artifacts that we can deploy them as `Extension` in the future (see https://github.com/gardener/gardener/issues/9635).

I had to adapt the specification of `image` in `gardener-extension-shoot-lakom-service` Helm chart that the pipeline is able to overwrite the values correctly.

Fixes #

**Release note**:

```feature operator
Helm charts of extension and admission controller are published as OCI artifacts now.
```
```breaking operator
The specification of the `image` in `gardener-extension-shoot-lakom-service` Helm chart has been changed.
```


charts/gardener-extension-shoot-lakom-service/templates/deployment.yaml
```diff
@@ -69,7 +69,8 @@ spec:
           type: RuntimeDefault
       containers:
       - name: gardener-extension-{{ .Values.serviceName }}
-        image: {{ .Values.image }}
+        image: {{ include "image" . }}
```
vpnachev, line 72: `image` template needs to be defined.

vpnachev (changes requested):
Some preliminary feedback, I will need some time for the skaffold/kustomize changes.

charts/gardener-extension-shoot-lakom-service/templates/deployment.yaml
oliver-goetz: It is already existing, isn't it?
https://github.com/gardener/gardener-extension-shoot-lakom-service/blob/0adc6bb07ca55f5fe90278cf681063f0cd2553ef/charts/gardener-extension-shoot-lakom-service/templates/_helpers.tpl#L1-L7

.gitignore
```diff
@@ -20,6 +20,5 @@ testbin
 *~
 dive.log
 /tmp
-/example/lakom
```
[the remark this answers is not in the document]
oliver-goetz: I did not notice that it is used by `make dev-setup` too and reverted the change.

vpnachev (approved):
/lgtm

PR #99, opened 2024-07-26
Add support for RSASSA-PSS scheme

vpnachev:
Add support for RSASSA-PSS scheme

/area ipcei
/kind enhancement

Fixes #

Draft until proper release notes are written. 

**Release note**:

```breaking operator
:warning: Lakom admission controller has removed the flag `--cosign-public-key-path`.
The trusted public keys are now configurable via the flag `--lakom-config-path` as the file is no longer bundle of public keys, but a structured configuration in YAML or JSON format.
Now each key must be named and optionally a signature verification algorithm can be specified. For more details, check the [lakom usage](https://github.com/gardener/gardener-extension-shoot-lakom-service/blob/main/docs/usage/lakom.md) page.
```

```breaking operator
:warning: The `lakom` helm chart value `cosign.publicKeys` changed its semantic, it is no longer list of keys, but list of tuple of `name`, `key` and optionally `algorithm`.
```

```feature operator
Lakom adds support for `RSASSA-PSS` scheme for signature verification next to `RSASSA-PKCS1-v1_5`, it can be configured via the `algorithm` field associated with each public key. The algorithm field also provides control over the hash func to be used.
```

```breaking operator
:warning: Lakom extension controller helm chart values "controllers.cosignPublicKeys" changed its semantic, it is no longer list of keys, but list of tuple of `name`, `key` and optionally `algorithm`. The same change is applied to the extension configuration file field `cosignPublicKeys`.
```


pkg/lakom/config/config.go
```diff
@@ -0,0 +1,129 @@
...
+
+func parseRSAAlgorithm(algorithm AlgorithmKey) (*crypto.Hash, *RSASchemeKey, error) {
+	switch algorithm {
+	case "":
+		// RSASSA-PKCS1-v1_5-SHA256 is the default algorithm
+		fallthrough
+
+	// RSASSA-PKCS1-v1_5
+	case RSAPKCS1v15SHA256:
+		return ptr.To(crypto.SHA256), ptr.To(RSAPKCS1v15), nil
```
rrhubenov, line 112: What's the advantage of using `ptr.To` instead of just returning address with `&`?

Also, are the Hash and RSASchemeKey pointers only for optimization purposes?

pkg/apis/config/v1alpha1/types.go
```diff
@@ -20,7 +21,8 @@ type Configuration struct {
 	// +optional
 	HealthCheckConfig *healthcheckconfigv1alpha1.HealthCheckConfig `json:"healthCheckConfig,omitempty"`
 	// CosignPublicKeys is the cosign public keys used to verify image signatures.
-	CosignPublicKeys []string `json:"cosignPublicKeys,omitempty"`
+	// +optional
+	CosignPublicKeys *runtime.RawExtension `json:"cosignPublicKeys,omitempty"`
```
rrhubenov, line 25: I'm unsure why this can't be typed to a specific struct?

pkg/apis/config/types.go
```diff
@@ -18,7 +19,7 @@ type Configuration struct {
 	// HealthCheckConfig is the config for the health check controller.
 	HealthCheckConfig *healthcheckconfig.HealthCheckConfig
 	// CosignPublicKeys is the cosign public keys used to verify image signatures.
-	CosignPublicKeys []string
+	CosignPublicKeys *runtime.RawExtension
```
rrhubenov, line 22: Looking at the docs for RawExtension[1], it's advised that the external type is runtime.RawExtension while the internal one is runtime.Object. Would this be applicable in this case?

[1] https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime#RawExtension

pkg/controller/seed/reconciler.go
```diff
@@ -103,10 +103,15 @@ func (kcr *kubeSystemReconciler) reconcile(ctx context.Context, logger logr.Logg
...
+	lakomConfig, err := yaml.JSONToYAML(kcr.serviceConfig.CosignPublicKeys.Raw)
+	if err != nil {
+		return fmt.Errorf("failed to convert lakom config from json to yaml, %w", err)
+	}
+
 	resources, err := getResources(
 		generatedSecrets[constants.SeedWebhookTLSSecretName].Name,
 		image.String(),
-		kcr.serviceConfig.CosignPublicKeys,
+		string(lakomConfig),
```
dimityrmirchev, line 114: You can directly pass this as `[]byte` and then use `.secret.data` instead of `.secret.stringData`
```suggestion
		lakomConfig,
```

rrhubenov (commented):
Looks nice, I have just a few questions.

Also, could you give me some steps to test this manually on my machine?

pkg/lakom/config/types.go
```diff
@@ -0,0 +1,75 @@
...
+// that should be used during signature verification.
+type VerifierKey struct {
+	// Name is a human readable name associated with the key.
+	Name string
+	// Key is the public crypto key.
+	Key crypto.PublicKey
+	// Hash is the hash function that the verifier should use during signature verification.
+	Hash *crypto.Hash
+	// Scheme is the RSA scheme (if the key is of RSA type) the verifier to use during signature verification.
+	Scheme *RSASchemeKey
```
dimityrmirchev, line 74: If this scheme is only RSA related I think prefixing the name with `RSA` makes it easier to understand
```suggestion
	RSAScheme *RSASchemeKey
```

dimityrmirchev (commented):
In addition to @rrhubenov questions two small comments from my side.

pkg/lakom/config/config.go
vpnachev: It is not allowed to get the address of a constant, see https://go.dev/ref/spec#Address_operators
```bash
$ make start-lakom
...
# github.com/gardener/gardener-extension-shoot-lakom-service/pkg/lakom/config
pkg/lakom/config/config.go:112:11: invalid operation: cannot take address of crypto.SHA256 (constant 5 of type crypto.Hash)
pkg/lakom/config/config.go:112:27: invalid operation: cannot take address of RSAPKCS1v15 (constant "RSASSA-PKCS1-v1_5" of type RSASchemeKey)
```

pkg/apis/config/v1alpha1/types.go
vpnachev: Lakom extension controller config should not be coupled to the lakom admission controller config, a good practice is to embed such configurations as `runtime.RawExtension`.

pkg/apis/config/types.go
vpnachev: I am aware of it, and even tried to implement it this way, however I decided not to follow this recommendation because of several reasons

- these types are not real API objects exposed via REST server, they are just used to introduce some well-defined structure of the configuration file
- the external version type is practically not used anywhere in code, it is indirectly used when the config file is read and decoded into the intern version  https://github.com/gardener/gardener-extension-shoot-lakom-service/blob/5de099d748625180681d396531d7f8d3bfb02b7f/pkg/cmd/options.go#L62-L63 
-  if the internal type is using `runtime.Object`, the lakom admission controller config type must follow the same path and be like a k8s API type with conversions, deep copies, defaulting, API scheme, API group, etc. I decided to save this overhead and use a simple go type that can be decoded/encoded from/to yaml file( also json).

pkg/lakom/config/config.go
rrhubenov: Thanks! Did not know that. It does make sense.

pkg/apis/config/types.go
rrhubenov: Makes sense 👍

pkg/controller/seed/reconciler.go
vpnachev: With cfb8298 the lakom config is now stored in ConfigMap instead of Secret, now `lakomConfig` should be of type string wherever possible.

dimityrmirchev (approved):
/lgtm

PR #103, opened 2024-08-07
Scope based verification

rrhubenov:
Enables a lakom user to choose the scope that lakom verifies.

Fixes #

**Release note**:

```feature user
You can now choose the scope that Lakom verifies. Previously, the scope was always the pods in the kube-system namespace that are annotated with `managed-by/gardener`. 
```


charts/admission/charts/application/Chart.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+description: A Helm chart to deploy the gardener-extension-registry-cache-admission application related resources
```
rrhubenov, line 2: Replace with 'gardener-extension-shoot-lakom'

charts/admission/charts/runtime/Chart.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+description: A Helm chart to deploy the gardener-extension-registry-cache-admission runtime related resources
```
rrhubenov, line 2: replace with 'gardener-extension-shoot-lakom-admission'

pkg/admission/validator/lakom/shoot.go
```diff
@@ -0,0 +1,85 @@
...
+// The second return argument is the extension itself. An empty extension is returned if the extension is not found.
+func findExtension(extensions []core.Extension, extensionType string) (int, core.Extension) {
+	for i, ext := range extensions {
+		if ext.Type == extensionType {
+			return i, ext
+		}
+	}
+
+	return -1, core.Extension{}
+}
```
rrhubenov, line 45: Could I use some util from gardener/gardener? Seems like this should be a common task for all extensions that have a providerConfig in the shoot

pkg/admission/validator/lakom/shoot.go
```diff
@@ -0,0 +1,85 @@
...
+
+	// TODO(rrhubenov): I don't like this. This can easily be forgotten when a new scope is added.
+	switch lakomConfig.Scope {
+	case lakom.Cluster:
+	case lakom.KubeSystem:
+	case lakom.KubeSystemManagedByGardener:
+	default:
+		// TODO(rrhubenov): This logs the error but does not print to stderr of the client. Should be improved
+		allErrs = append(allErrs, field.Invalid(providerConfigPath.Child("scope"), lakomConfig.Scope, fmt.Sprintf("Invalid scope %s. Please refer to the documentation for available scopes", lakomConfig.Scope)))
+	}
```
rrhubenov, line 82: Imo, this should be refactored into an array that contains the available values and the array should be next to the definition of the constants in pkg/apis/lakom/types.go

pkg/apis/lakom/types.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ScopeType determines the namespaces and labels that will be monitored by lakom webhooks
+type ScopeType string
+
+const (
+	// KubeSystem denotes the `kube-system` namespace
+	KubeSystem ScopeType = "kubeSystem"
```
vpnachev, line 19: Use pascal case for the enumerated values, ref: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#constants

```suggestion
	KubeSystem ScopeType = "KubeSystem"
```

vpnachev (commented):
I will need to do at least one more round of review, currently focused mostly on naming and helm charts.

cmd/gardener-extension-shoot-lakom-service/app/app.go
```diff
@@ -116,6 +118,12 @@ func (o *Options) run(ctx context.Context) error {
 	if err := monitoringv1.AddToScheme(mgr.GetScheme()); err != nil {
 		return fmt.Errorf("could not update manager scheme: %w", err)
 	}
+	if err := v1alpha1apilakom.AddToScheme(mgr.GetScheme()); err != nil {
+		return fmt.Errorf("could not update manager scheme: %w", err)
+	}
+	if err := apilakom.AddToScheme(mgr.GetScheme()); err != nil {
+		return fmt.Errorf("could not update manager scheme: %w", err)
```
vpnachev, line 125: Update the error messages so that they distinguish when different APIs are added to the scheme.

hack/check-skaffold-deps.sh
```diff
@@ -20,8 +20,9 @@ function run() {
 }
 
 # skaffold.yaml
-run "skaffold.yaml" "gardener-extension-shoot-lakom-service"      "extension"
-run "skaffold.yaml" "lakom"                                       "extension"
+run "skaffold.yaml" "gardener-extension-shoot-lakom-service"                "extension"
+run "skaffold.yaml" "lakom"                                                 "extension"
+run "skaffold.yaml" "gardener-extension-shoot-lakom-service-admission"      "admission"
```
vpnachev, line 25: The binary name have to be adapted here, too.

pkg/admission/validator/lakom/lakom_suite_test.go
```diff
@@ -0,0 +1,130 @@
+package lakom_test
```
vpnachev, line 1: License header has to be added

pkg/admission/validator/lakom/lakom_suite_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"testing"
+
+	lakom "github.com/gardener/gardener-extension-shoot-lakom-service/pkg/admission/validator/lakom"
+	apilakom "github.com/gardener/gardener-extension-shoot-lakom-service/pkg/apis/lakom"
+	v1alpha1 "github.com/gardener/gardener-extension-shoot-lakom-service/pkg/apis/lakom/v1alpha1"
```
vpnachev, line 10:
```suggestion
	"github.com/gardener/gardener-extension-shoot-lakom-service/pkg/apis/lakom/v1alpha1"
```

pkg/admission/validator/lakom/lakom_suite_test.go
```diff
@@ -0,0 +1,130 @@
+package lakom_test
+
+import (
+	"context"
+	"encoding/json"
+	"testing"
+
+	lakom "github.com/gardener/gardener-extension-shoot-lakom-service/pkg/admission/validator/lakom"
```
vpnachev, line 8:
```suggestion
	"github.com/gardener/gardener-extension-shoot-lakom-service/pkg/admission/validator/lakom"
```

pkg/admission/validator/lakom/lakom_suite_test.go
```diff
@@ -0,0 +1,130 @@
...
+						},
+					},
+				},
+			}
+		})
+
+		It("should return err when new is not a Shoot", func() {
+			err := shootValidator.Validate(ctx, &corev1.Pod{}, nil)
+
+			Expect(err).To(HaveOccurred())
```
vpnachev, line 82:
```suggestion
			Expect(shootValidator.Validate(ctx, &corev1.Pod{}, nil)
).To(HaveOccurred())
```

You can add a check that the return error is the expected one.

pkg/admission/validator/lakom/lakom_suite_test.go
```diff
@@ -0,0 +1,130 @@
...
+				Raw: encode(&v1alpha1.LakomConfig{
+					TypeMeta: metav1.TypeMeta{
+						APIVersion: v1alpha1.SchemeGroupVersion.String(),
+						Kind:       "LakomConfig",
+					},
+					Scope: "invalid",
+				}),
+			}
+
+			shoot.Spec.Extensions[0].ProviderConfig = extension
```
vpnachev, line 115:
```suggestion
			shoot.Spec.Extensions[0].ProviderConfig = &runtime.RawExtension{
				Raw: encode(&v1alpha1.LakomConfig{
					TypeMeta: metav1.TypeMeta{
						APIVersion: v1alpha1.SchemeGroupVersion.String(),
						Kind:       "LakomConfig",
					},
					Scope: "invalid",
				}),
			}
```

pkg/admission/validator/lakom/lakom_suite_test.go
```diff
@@ -0,0 +1,130 @@
...
+				"Type":  Equal(field.ErrorTypeInvalid),
+				"Field": Equal("spec.extensions[0].providerConfig.scope"),
+			}))))
+		})
+
+	})
+})
+
+func encode(obj runtime.Object) []byte {
+	data, _ := json.Marshal(obj)
```
vpnachev, line 128: This error should not be ignored. Maybe you can avoid using the `encode` function by setting Raw directly to the byte array representation of a string 

```go
	extension := &runtime.RawExtension{
		Raw: []byte(`apiVersion: lakom.extensions.gardener.cloud/v1alpha1
kind: LakomConfig
scope: invalid
`),
```

pkg/admission/validator/lakom/shoot.go
```diff
@@ -0,0 +1,82 @@
...
+	if !ok {
+		return fmt.Errorf("wrong object type %T", new)
+	}
+
+	i, lakomExt := findExtension(shoot.Spec.Extensions, constants.ExtensionType)
+	if i == -1 {
+		return nil
+	}
+
+	providerConfigPath := field.NewPath("spec", "extensions").Index(i).Child("providerConfig")
```
vpnachev, line 69: Define the variable later in the function closer to the place where it will be used.

pkg/admission/validator/lakom/shoot.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	lakomConfig := &lakom.LakomConfig{}
+	if err := runtime.DecodeInto(s.decoder, lakomExt.ProviderConfig.Raw, lakomConfig); err != nil {
+		return fmt.Errorf("failed to decode providerConfig: %w", err)
+	}
+
+	allErrs = append(allErrs, s.validateScopeType(providerConfigPath.Child("scope"), lakomConfig.Scope)...)
+
+	return allErrs.ToAggregate()
```
vpnachev, line 81:
```suggestion
	return s.validateScopeType(providerConfigPath.Child("scope"), lakomConfig.Scope).ToAggregate()
```

pkg/admission/validator/lakom/webhook.go
```diff
@@ -0,0 +1,44 @@
...
+func New(mgr manager.Manager) (*extensionswebhook.Webhook, error) {
+	logger.Info("Setting up webhook", "name", Name)
+
+	decoder := serializer.NewCodecFactory(mgr.GetScheme(), serializer.EnableStrict).UniversalDecoder()
+	apiReader := mgr.GetAPIReader()
+
+	return extensionswebhook.New(mgr, extensionswebhook.Args{
+		Provider: constants.ExtensionType,
+		Name:     Name,
+		Path:     "/webhooks/shoot-lakom-service",
```
vpnachev, line 35:
```suggestion
		Path:     "/webhooks/shoot-lakom-validator",
```

or 

```suggestion
		Path:     "/webhooks/shoot-lakom-admission",
```

vpnachev (commented):
I am now thinking that the program name should be `gardener-admission-shoot-lakom-admission`, not `gardener-extension-shoot-lakom-admission` (similar to gardener-admission-provider-aws).

pkg/admission/validator/lakom/lakom_suite_test.go
rrhubenov: I do like that with this encode we can leverage the type system to avoid stupid mistakes with spelling at compile time. I'd prefer if we just handle the error. This error isn't something that we're testing and this code being just a unit test, are you okay with just panicking if there's an error?

pkg/admission/validator/lakom/shoot.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	return -1, core.Extension{}
+}
+
+func (s *shoot) validateScopeType(fldPath *field.Path, scopeType lakom.ScopeType) field.ErrorList {
+	errList := field.ErrorList{}
+
+	if !lakom.AllowedScopes.Has(scopeType) {
+		errList = append(errList, field.Invalid(fldPath, scopeType, fmt.Sprintf("Invalid scope %s. Please refer to the documentation for available scopes", scopeType)))
```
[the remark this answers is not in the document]
rrhubenov: I agree, thanks!

pkg/controller/lifecycle/actuator.go
```diff
@@ -104,6 +105,17 @@ func (a *actuator) Reconcile(ctx context.Context, logger logr.Logger, ex *extens
 		return err
 	}
 
+	if ex.Spec.ProviderConfig == nil {
+		return fmt.Errorf(".spec.providerConfig is required for the lakom extension")
+	}
```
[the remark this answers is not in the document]
rrhubenov: Oh, yeah. I've actually handled the case in the decode step but I guess forgot to remove this check. Thanks!

rrhubenov:
@vpnachev Thanks for the thorough review! There's just one thing I've left unresolved that we can discuss. 

Also, the CI pipeline shows that golangci-lint is panicking on all the linters and thats why the it's failing. Locally, using the same version of golangci-lint, there doesn't seem to be such an issue. Any ideas?

docs/usage/shoot-extension.md
```diff
@@ -13,5 +13,19 @@ spec:
   extensions:
   - type: shoot-lakom-service
     disabled: true
+    providerConfig:
+      apiVersion: lakom.extensions.gardener.cloud/v1alpha1
+      kind: LakomConfig
+      scope: kubeSystem
```
vpnachev, line 19:
```suggestion
      scope: KubeSystem
```

docs/usage/shoot-extension.md
```diff
@@ -13,5 +13,19 @@ spec:
...
 ```
+
+The `scope` field instruct lakom which pods to validate. The possible values are:
+
+- `kubeSystem`
+Lakom will validate all pods in the `kube-system` namespace.
+- `kubeSystemManagedByGardener`
+Lakom will validate all pods in the `kube-system` namespace that are annotated with "managed-by/gardener"
+- `cluster`
+Lakom will validate all pods in all namespaces.
```
vpnachev, line 30:
```suggestion
- `KubeSystem`
Lakom will validate all pods in the `kube-system` namespace.
- `KubeSystemManagedByGardener`
Lakom will validate all pods in the `kube-system` namespace that are annotated with "managed-by/gardener"
- `Cluster`
Lakom will validate all pods in all namespaces.
```

pkg/admission/validator/lakom/lakom_suite_test.go
vpnachev: You can also add check that the error have not occurred here

```suggestion
	data, err := json.Marshal(obj)
	Expect(err).ToNot(HaveOccurred())
```

pkg/admission/validator/lakom/shoot.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	return errList
+}
+
+// Validate validates the given shoot object
+func (s *shoot) Validate(_ context.Context, new, _ client.Object) error {
+	shoot, ok := new.(*core.Shoot)
+	if !ok {
+		return fmt.Errorf("wrong object type %T", new)
```
vpnachev, line 59:
```suggestion
		return fmt.Errorf("wrong object type %T, expected core.Shoot", new)
```

This should make the error message more helpful.

pkg/admission/validator/lakom/shoot.go
```diff
@@ -0,0 +1,81 @@
...
+	lakomConfig := &lakom.LakomConfig{}
+	if err := runtime.DecodeInto(s.decoder, lakomExt.ProviderConfig.Raw, lakomConfig); err != nil {
+		return fmt.Errorf("failed to decode providerConfig: %w", err)
+	}
+	if lakomConfig.Scope == nil {
+		return nil
+	}
+
+	providerConfigPath := field.NewPath("spec", "extensions").Index(i).Child("providerConfig")
+	if lakomExt.ProviderConfig == nil {
```
vpnachev, line 76: This check here is a bit odd and should be never true. The reason is that a few lines earlier `lakomExt.ProviderConfig.Raw` is used and it would panic if `lakomExt.ProviderConfig` is nil.

The check still makes sense, it just have to be earlier in the function.

pkg/apis/lakom/v1alpha1/types.go
```diff
@@ -0,0 +1,24 @@
...
+
+package v1alpha1
+
+import (
+	lakom "github.com/gardener/gardener-extension-shoot-lakom-service/pkg/apis/lakom"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +genclient
```
vpnachev, line 13:
```suggestion
```

This resource will not be served by the API, therefore a client is not needed.

pkg/apis/lakom/v1alpha1/types.go
```diff
@@ -0,0 +1,24 @@
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
```
vpnachev, line 15:
```suggestion
// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
```

Duplicated line

pkg/apis/lakom/register.go
```diff
@@ -0,0 +1,41 @@
...
+
+package lakom
+
+import (
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// GroupName is the group name use in this package
+const GroupName = "lakom.extensions.config.gardener.cloud"
```
vpnachev, line 13:
```suggestion
const GroupName = "lakom.extensions.gardener.cloud"
```

Please, update all other occurrences where the group name with `.config.` used.

pkg/apis/lakom/types.go
```diff
@@ -0,0 +1,37 @@
...
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ScopeType determines the namespaces and labels that will be monitored by lakom webhooks
+type ScopeType string
+
+const (
+	// KubeSystem denotes the `kube-system` namespace
+	KubeSystem ScopeType = "KubeSystem"
+	// KubeSystemManagedByGardener denoes pods in the `kube-system` namespace that have a `managed-by/gardener` label
```
vpnachev, line 20:
```suggestion
	// KubeSystemManagedByGardener denotes pods in the `kube-system` namespace that have a `managed-by/gardener` label
```

pkg/apis/lakom/types.go
```diff
@@ -0,0 +1,37 @@
...
+	KubeSystemManagedByGardener ScopeType = "KubeSystemManagedByGardener"
+	// Cluster denotes the whole cluster
+	Cluster ScopeType = "Cluster"
+)
+
+// AllowedScopes lists the scopes that can be chosen for lakom
```
vpnachev, line 26:
```suggestion
// AllowedScopes lists the scopes that can be chosen for lakom.
```

pkg/controller/lifecycle/actuator_test.go
```diff
@@ -49,49 +51,69 @@ var _ = Describe("Actuator", func() {
...
-			extensionNamespace            = "shoot--foo--bar"
-			shootAccessServiceAccountName = "extension-shoot-lakom-service-access"
+			shootNamespace                  = "garden-foo"
+			extensionNamespace              = "shoot--foo--bar"
+			scope                           = lakom.KubeSystemManagedByGardener
+			shootAccessServiceAccountName   = "extension-shoot-lakom-service-access"
+			validatingWebhookKey            = "validatingwebhookconfiguration____gardener-extension-shoot-lakom-service-shoot.yaml"
+			mutatingWebhookKey              = "mutatingwebhookconfiguration____gardener-extension-shoot-lakom-service-shoot.yaml"
+			roleKey                         = "role__kube-system__gardener-extension-shoot-lakom-service-resource-reader.yaml"
+			roleBindingKey                  = "rolebinding__kube-system__gardener-extension-shoot-lakom-service-resource-reader.yaml"
```
vpnachev, line 61: This test was reworked with #109 as the rendered managed resources are now compressed with Brotli and put into single manifest, hence the secret data keys are not used anymore.  I guess something went wrong when the conflicts were resolved after the main branch was merged into the PR branch.

pkg/controller/lifecycle/actuator_test.go
```diff
@@ -126,6 +148,23 @@ var _ = Describe("Actuator", func() {
...
+				Expect(err).ToNot(HaveOccurred())
+				manifests, err := test.ExtractManifestsFromManagedResourceData(resources)
+				Expect(err).ToNot(HaveOccurred())
+
+				Expect(manifests).To(ContainElements(
+					expectedShootMutatingWebhook(caBundle, extensionNamespace, objectSelector, namespaceSelector),
+					expectedSeedValidatingWebhook(caBundle, extensionNamespace, objectSelector, namespaceSelector),
+				))
+			},
+			Entry("KubeSystem managed by Gardener scope", lakom.KubeSystemManagedByGardener, managedByGardenerObjectSelector, kubeSystemNamespaceSelector),
```
vpnachev, line 163:
```suggestion
			Entry("KubeSystemManagedByGardener scope", lakom.KubeSystemManagedByGardener, managedByGardenerObjectSelector, kubeSystemNamespaceSelector),
```

pkg/controller/lifecycle/actuator.go
```diff
@@ -104,6 +105,17 @@ func (a *actuator) Reconcile(ctx context.Context, logger logr.Logger, ex *extens
 		return err
 	}
 
+	lakomProviderConfig := &lakom.LakomConfig{}
+	if _, _, err := a.decoder.Decode(ex.Spec.ProviderConfig.Raw, nil, lakomProviderConfig); err != nil {
+		// Apply default values if provider config has not been provided
+		logger.Error(err, "Could not decode provider config. Using default value `kubeSystemManagedByGardener` for scope")
```
vpnachev, line 111:
```suggestion
		logger.Error(err, "Could not decode provider config. Using default value `KubeSystemManagedByGardener` for scope")
```

pkg/controller/lifecycle/actuator.go
```diff
@@ -104,6 +105,17 @@ func (a *actuator) Reconcile(ctx context.Context, logger logr.Logger, ex *extens
...
+		lakomProviderConfig.Scope = ptr.To(lakom.KubeSystemManagedByGardener)
+	}
+	if lakomProviderConfig.Scope == nil {
+		logger.Info("No scope specified. Using default value `kubeSystemManagedByGardener` for scope")
```
vpnachev, line 115:
```suggestion
		logger.Info("No scope specified. Using default value `KubeSystemManagedByGardener` for scope")
```

vpnachev (commented):
Thanks!

I think I have found just nitpicks.

pkg/admission/validator/lakom/lakom_suite_test.go
rrhubenov: I try to avoid testing for things that are outside the scope of the unit we're writing the tests for. Especially if the only way I see this could fail is if there's a bug somewhere in the `json` lib, in its dependencies, or we've passed something malformed somehow.

Either way, I don't have a super strong opinion on this and I'm interested to see if you have some opinions on this topic.

vpnachev: We are more interested that the `obj` could be successfully marshaled, it is not expected to catch bugs in json.Marshal. Just ignoring the error is not correct because then we cannot rely the value in `data` variable is correct. 

I am not sure how panic will be helpful here, I guess it could interrupt the execution of the other tests and might not provide meaningful/helpful information what was the input when the panic occurred. 

To sum up, I am fine just to ensure that there `json.Marshal` does not return error.

charts/gardener-extension-shoot-lakom-admission/charts/runtime/templates/rbac.yaml
```diff
@@ -0,0 +1,58 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ include "name" . }}
+  namespace: garden
```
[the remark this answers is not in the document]
rrhubenov: This chart has been reworked from the chart that is used for the `registry-cache`  extension, so I'm not completely sure why the role is for the garden namespace. 

I don't see any self-bootstrapping functionality on the admission controller, so it might be something specific for the `registry-cache` that can be safely removed for the lakom admission controller. Another possibility is that the `registry-cache` also borrowed this chart from some other extension. I'll ask them and get back to you

cmd/gardener-extension-shoot-lakom-admission/app/app.go
```diff
@@ -0,0 +1,174 @@
...
+			mgr, err := manager.New(restOpts.Completed().Config, managerOptions)
+			if err != nil {
+				return fmt.Errorf("failed to instantiate manager: %w", err)
+			}
+
+			if err := v1alpha1apilakom.AddToScheme(mgr.GetScheme()); err != nil {
+				return fmt.Errorf("failed to add `lakom.extensions.gardener.cloud/v1alpha1` to manager scheme: %w", err)
+			}
+			if err := apilakom.AddToScheme(mgr.GetScheme()); err != nil {
+				return fmt.Errorf("failed to add internal version of `lakom.extensions.config.gardener.cloud` to manager scheme: %w", err)
```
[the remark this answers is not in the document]
rrhubenov: Not at all. I think this was some copy-pasting issue on my part. Thank you for catching

charts/gardener-extension-shoot-lakom-admission/charts/runtime/templates/rbac.yaml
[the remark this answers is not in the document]
rrhubenov: @vpnachev I'll be resolving since it seem like we'll have to talk with Tim Usner about this. Based on the convo with Ismail, it's best we leave the namespace as `garden` for now.

vpnachev (approved):
/lgtm

skaffold.yaml
```diff
@@ -90,3 +92,34 @@ manifests:
...
+deploy:
+  helm:
+    releases:
+      - name: gardener-extension-shoot-lakom-service-extension
+        namespace: garden
+        wait: true
+        chartPath: charts/admission
+        setValueTemplates:
+          global.image.repository: '{{.IMAGE_REPO_local_skaffold_gardener_extension_shoot_lakom_service_admission}}'
+          global.image.tag: '{{.IMAGE_TAG_local_skaffold_gardener_extension_shoot_lakom_service_admission}}@{{.IMAGE_DIGEST_local_skaffold_gardener_extension_shoot_lakom_service_admission}}'
```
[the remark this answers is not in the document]
rrhubenov: Thank you!

pkg/controller/lifecycle/actuator.go
```diff
@@ -104,6 +105,17 @@ func (a *actuator) Reconcile(ctx context.Context, logger logr.Logger, ex *extens
 		return err
 	}
 
+	lakomProviderConfig := &lakom.LakomConfig{}
+	if _, _, err := a.decoder.Decode(ex.Spec.ProviderConfig.Raw, nil, lakomProviderConfig); err != nil {
+		// Apply default values if provider config has not been provided
+		logger.Error(err, "Could not decode provider config. Using default value `KubeSystemManagedByGardener` for scope")
+		lakomProviderConfig.Scope = ptr.To(lakom.KubeSystemManagedByGardener)
+	}
```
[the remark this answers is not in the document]
rrhubenov: Huh, I would have expected that Decode return a non-nil error if the provided data slice is nil.
This pattern is also used in the registry, so I can add a PR to fix this there as well. Nice catch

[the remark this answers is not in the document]
vpnachev: The issue happens just before the decoder.Decode is called when its parameters are evaluated and then trying to read a field from a nil pointer results in the segmentation fault. If you see the same in other places, it will be very nice if you open PRs to fix them.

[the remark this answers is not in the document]
rrhubenov: Oh, yes, the actual nil field is the ProviderConfig, my bad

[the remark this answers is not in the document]
rrhubenov: Now that I understood correctly, I can see that they've handled the nil ProviderConfig case in the registry as well, so it was an issue just here

PR #106, opened 2024-08-20
Disable lakom via annotation from seed

rrhubenov:
ManagedSeed clusters do not need a lakom deployment since they can already be verified by the seed that's managing them. This feature allows us to configure this using an annotation on the seed resources of the managedseed.
Fixes #

**Release note**:

```feature operator
It is now possible to disable installation of lakom admission webhook in the `kube-system` namespace of seed clusters. This is useful for managed seeds that have the lakom extension enabled with `Cluster` scope. To disable the installation, the seed must be annotated with `service.lakom.extensions.gardener.cloud/enable-lakom-admission-controller=false`
```


charts/gardener-extension-shoot-lakom-service/values.yaml
```diff
@@ -59,3 +59,8 @@ controllers:
   debugConfig:
     enableProfiling: false
     enableContentionProfiling: false
+
+gardener:
+  seed:
+    annotations:
+      service.lakom.extensions.gardener.cloud/deploySeedResources: true
```
vpnachev, line 66: Is it necessary to set these helm values here? They will be considered as default values when the chart is installed. 

 
Also, the annotation key should not be camelCase, see https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#label-selector-and-annotation-conventions

> Key names should be all lowercase, with words separated by dashes instead of camelCase

For annotation key I would recommend `seed.lakom.extensions.gardener.cloud/enable-admission-controller`

pkg/apis/config/types.go
```diff
@@ -33,6 +33,8 @@ type Configuration struct {
 	// AllowInsecureRegistries sets the lakom webhook to allow HTTP communication with OCI registries.
 	// It first tries HTTPS and then falls back to HTTP.
 	AllowInsecureRegistries bool
+	// DeploySeedResources instruct the lakom-controller whether it should deploy lakom to the seed cluster.
+	DeploySeedResources bool
```
vpnachev, line 37: Move this setting to `SeedBootstrap` type.  Then you can rename the field to `Enabled`

pkg/controller/seed/reconciler.go
```diff
@@ -80,6 +80,16 @@ func (kcr *kubeSystemReconciler) reconcile(ctx context.Context, logger logr.Logg
 	const (
 		kubeSystemNamespaceName = metav1.NamespaceSystem
 	)
+	var (
+		ownerNamespace = kcr.serviceConfig.SeedBootstrap.OwnerNamespace
+	)
+
+	if !kcr.serviceConfig.DeploySeedResources {
+		if err := managedresources.DeleteForSeed(ctx, kcr.client, ownerNamespace, constants.ManagedResourceNamesSeed); err != nil {
```
vpnachev, line 88: Let's add log messages showing when the admission webhook is being uninstalled, and when installed.

```suggestion
	if !kcr.serviceConfig.DeploySeedResources {
		logger.Info("Deleting lakom admission controller from the seed cluster")
		if err := managedresources.DeleteForSeed(ctx, kcr.client, ownerNamespace, constants.ManagedResourceNamesSeed); err != nil {
```

pkg/controller/seed/reconciler.go
```diff
@@ -80,6 +80,16 @@ func (kcr *kubeSystemReconciler) reconcile(ctx context.Context, logger logr.Logg
...
+			return err
+		}
+		return nil
+	}
```
vpnachev, line 92:
```suggestion
	}

	logger.Info("Installing lakom admission controller from the seed cluster")
```

vpnachev (changes requested):
Thanks!
I have some change requests in the in-code comments

charts/gardener-extension-shoot-lakom-service/values.yaml
rrhubenov: Wouldn't we want them exactly for using them as default values when the annotation is not added?

rrhubenov: > For annotation key I would recommend seed.lakom.extensions.gardener.cloud/enable-admission-controller
My idea with the naming was to make it clear that the controller is still enabled but it just won't deploy the resources monitoring and validating the images on the seed.

Does `seed.lakom.extensions.gardener.cloud/enable-admission-controller` make the same thing clear? I don't want someone to interpret the option as completely disabling the controller.

rrhubenov: > Not for values that gardener manages. The values files are also used as reference document what options are available and we should not expose what gardener is additionally injecting.
Thanks, I hadn't thought about this.

> but it can be improved a bit more if the annotation key is named seed.lakom.extensions.gardener.cloud/enable-lakom-admission-controller
I agree, I'll use this.

rrhubenov: Actually, I'll hold out on making the change for now because of the discussion about deciding whether we're on a managedseed based on the existance of a `validatingwebhookconfiguration`.

charts/gardener-extension-shoot-lakom-service/templates/configmap-extension-config.yaml
```diff
@@ -19,6 +19,7 @@ data:
     useOnlyImagePullSecrets: {{ .Values.controllers.useOnlyImagePullSecrets }}
     allowUntrustedImages: {{ .Values.controllers.allowUntrustedImages }}
     allowInsecureRegistries: {{ .Values.controllers.allowInsecureRegistries }}
+    deploySeedResources: {{ get .Values.gardener.seed.annotations "service.lakom.extensions.gardener.cloud/deploySeedResources" }}
```
vpnachev, line 22: The example needs to be aligned with the changes in the API.

```suggestion
      enabled: true
    useOnlyImagePullSecrets: {{ .Values.controllers.useOnlyImagePullSecrets }}
    allowUntrustedImages: {{ .Values.controllers.allowUntrustedImages }}
    allowInsecureRegistries: {{ .Values.controllers.allowInsecureRegistries }}
```

pkg/controller/seed/reconciler.go
```diff
@@ -80,6 +80,18 @@ func (kcr *kubeSystemReconciler) reconcile(ctx context.Context, logger logr.Logg
...
+	var (
+		ownerNamespace = kcr.serviceConfig.SeedBootstrap.OwnerNamespace
+	)
+
+	if !kcr.serviceConfig.SeedBootstrap.Enabled {
+		if err := managedresources.DeleteForSeed(ctx, kcr.client, ownerNamespace, constants.ManagedResourceNamesSeed); err != nil {
+                        logger.Info("Deleting lakom admission controller from the seed cluster")
+			return err
+		}
+		return nil
```
vpnachev, line 92:
```suggestion
                logger.Info("Deleting lakom admission controller from the seed cluster")
		return managedresources.DeleteForSeed(ctx, kcr.client, ownerNamespace, constants.ManagedResourceNamesSeed)
```

vpnachev (changes requested):

vpnachev (approved):
/lgtm

PR #109, opened 2024-09-18
Update github.com/gardener/gardener to v1.103.0

vpnachev:
Update github.com/gardener/gardener to v1.103.0

Fixes #

**Release note**:

```other dependency
The following third party dependencies have been updated:
- github.com/gardener/gardener v1.103.0
- github.com/google/go-containerregistry v0.20.0
- k8s.io/api v0.29.8
- k8s.io/apimachinery v0.29.8
- k8s.io/client-go v0.29.8
- k8s.io/code-generator v0.29.8
- k8s.io/component-base v0.29.8
- sigs.k8s.io/controller-runtime v0.17.6
```


pkg/controller/lifecycle/actuator_test.go
```diff
@@ -191,42 +178,35 @@ var _ = Describe("Actuator", func() {
...
+				manifests := strings.Split(string(data), "\n---\n") // Just '---\n' does not work because of the header/footer in the public keys that match the same manifest separator
+				Expect(manifests).To(HaveLen(7))
 
-					strResource := string(resource)
-					Expect(strResource).To(Equal(expectedResource), func() string {
-						return fmt.Sprintf("\nkey=%q\ngenerated resource\n%s\nexpectedResource\n%s\n", key, strResource, expectedResource)
-					})
+				for i := 0; i < len(manifests)-1; i++ { // Re-add the trailing '\n' removed during the split from the separator above
+					manifests[i] += "\n"
 				}
```
rrhubenov, line 193: Isn't
```
for i, _ := range manifests {
		manifests[i] += "\n"
}
```
considered more readable in Go? I've rarely seen the original for loops used for this case.
I would expect that the Go compiler can infer that the '_' variable isn't used and thus not make a copy of the object since it's not necessary.

rrhubenov (approved):
Looks good, just a small nitpick

pkg/controller/lifecycle/actuator_test.go
vpnachev: Yes, but the last element of the slice must not get the `\n` suffix. Then it should be

```go
for i := range manifests {
	if i == len(manifests)-1 {
		break
	}
	manifests[i] += "\n"
}
```

In this case I cannot say which is more readable or idiomatic. 
I am fine with both ways, what is your preference?

rrhubenov: My bad, didn't notice that the loop skips the last element. I'd much rather prefer your original for loop rather than the range in this case 👍

vpnachev: Ok, I however decided to change it to 
```go
for i := range manifests { // Re-add the trailing '\n' removed during the split from the separator above
	if i < len(manifests)-1 {
		manifests[i] += "\n"
	}
}
```

It should be now visible that the last element is skipped.

rrhubenov: I think we kind of overcomplicated the issue. I still prefer your original for loop but I'm find with this too 😃

rrhubenov (approved):

PR #110, opened 2024-09-21
Reduce resource demands of lakom container

vpnachev:
Reduce resource demands of lakom container as requested by @vlerenc:

> Given the stats below, can we please make the following changes:
> 
> - [ ] Memory requests changed from 64Mi to 25M as that's the 5th percentile of memory usage (generally used as initial requests unless there are good reasons to deviate)
> - [ ] CPU requests of 50m dropped (CPU usage is so low, that it actually isn't worth putting it under VPA and accepting the waste (VPA cannot recommend values below 10m/10M) and disruptions)
> - [ ] Vertical scaling on CPU dropped in VPA resource (see controlledResources ref1 and ref2) (consequently, we can/must drop CPU from VPA to keep CPU requests out of the pods)
> - [ ] minAllowed of 32Mi dropped (the average memory consumption is even lower than 32Mi, so we are wasting resources)

Fixes #

**Release note**:

```other operator
Lakom container resource demands have been reduced:
- memory requests reduced from 64Mi to 25M
- CPU requests have been dropped
- Vertical scaling on CPU requests dropped
- minAllowed of 32Mi memory dropped
```

### DIFF
--- a/charts/lakom/templates/vpa.yaml
+++ b/charts/lakom/templates/vpa.yaml
@@ -10,13 +10,11 @@ metadata:
   name: {{ .Values.name }}
   namespace: {{ .Release.Namespace }}
 spec:
-  {{- if .Values.vpa.resourcePolicy }}
   resourcePolicy:
     containerPolicies:
-    - containerName: '*'
-      minAllowed:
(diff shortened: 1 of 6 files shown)


rrhubenov (approved):
lgtm

Could you share where @vlerenc shared these comments?

vlerenc:
@rrhubenov Observations from some landscapes we could take a look at, but I have no permission to share.

PR #112, opened 2024-10-14
Allow use of client public keys

rrhubenov:
Currently, only the Gardener given public keys are used for verifying images for any pod.
This feature is necessary, if we want shoot owners to be able to verify their own
Fixes #

**Release note**:

```feature user
Public keys can now be added as a reference to a secret in the shoot spec.
```


rrhubenov:
I've tested this code and the resulting configmap looks like the following:
```
apiVersion: v1
data:
  config.yaml: |
    publicKeys:
    - algorithm: RSASSA-PKCS1-v1_5-SHA256
      key: |-
        -----BEGIN PUBLIC KEY-----
        MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEAKZef+D07q1AyXxf8xw3L23TgDSq
        Scitl0xDpdLGpODUZXG6LbWY3QXS6yhCt0MZH6+Jf5l/8daO7HDRRvbQMg==
        -----END PUBLIC KEY-----
      name: example-01
    - name: example-client-key1
      algorithm: RSASSA-PSS-SHA256
      key: |-
        -----BEGIN PUBLIC KEY-----
        MFwwDQYJKoZIhvcNAQEBBQADSwAwSAJBAPeQXbIWMMXYV+9+j9b4jXTflnpfwn4E
        GMrmqYVhm0sclXb2FPP5aV/NFH6SZdHDZcT8LCNsNgxzxV4N+UE/JIsCAwEAAQ==
        -----END PUBLIC KEY-----
    - name: example-client-key2
      algorithm: RSASSA-PSS-SHA256
      key: |-
        -----BEGIN PUBLIC KEY-----
        MFwwDQYJKoZIhvcNAQEBBQADSwAwSAJBAPeQXbIWMMXYV+9+j9b4jXTflnpfwn4E
        GMrmqYVhm0sclXb2FPP5aV/NFH6SZdHDZcT8LCNsNgxzxV4N+UE/JIsCAwEAAQ==
        -----END PUBLIC KEY-----
immutable: true
kind: ConfigMap
...
```

Where `example-client-key1` & `example-client-key2` were added as a resource to a secret with a `keys` field containing the said keys.

pkg/controller/lifecycle/actuator.go
```diff
@@ -153,18 +155,29 @@ func (a *actuator) Reconcile(ctx context.Context, logger logr.Logger, ex *extens
...
 	if err != nil {
 		return fmt.Errorf("failed to convert lakom config from json to yaml, %w", err)
 	}
 
+        var clientPublicKeys []byte
+	if lakomProviderConfig.PublicKeysSecretReference != nil {
+                var err error
+                clientPublicKeys, err = getClientKeys(ctx, a.client, cluster.Shoot.Spec.Resources, *lakomProviderConfig.PublicKeysSecretReference, namespace)
+                if err != nil {
+                    return fmt.Errorf("failed to find client public keys from the given reference %s", *lakomProviderConfig.PublicKeysSecretReference)
```
vpnachev, line 168:
```suggestion
                    return fmt.Errorf("failed to get the additional keys: %w", err)
```

It is important to wrap the original error as it will provide more information what actually failed down the stack, logging the resource reference name is kind of obsolete as it is well known and always can be read from the shoot resource.

pkg/controller/lifecycle/actuator.go
```diff
@@ -695,6 +708,31 @@ func getShootResources(webhookCaBundle []byte, extensionNamespace, shootAccessSe
 	return shootResources, nil
 }
 
+func getClientKeys(ctx context.Context, client client.Client, resources []v1beta1.NamedResourceReference, resourceName, namespace string) ([]byte, error) {
+    ref := v1beta1helper.GetResourceByName(resources, resourceName)
+    if ref == nil || ref.ResourceRef.Kind != "Secret" {
```
vpnachev, line 713: I will propose the split the condition and have dedicated error messages for both cases, one is that the configured `ResourceName` is missing in the `shoot.spec.resources`, the other is that it not of the expected kind `Secret`.

pkg/controller/lifecycle/actuator.go
```diff
@@ -695,6 +708,31 @@ func getShootResources(webhookCaBundle []byte, extensionNamespace, shootAccessSe
...
+
+    refSecret := &corev1.Secret{
+        ObjectMeta: metav1.ObjectMeta{
+            Name:      ref.ResourceRef.Name,
+            Namespace: namespace,
+        },
+    }
+
+    if err := controller.GetObjectByReference(ctx, client, &ref.ResourceRef, namespace, refSecret); err != nil {
+        return nil, fmt.Errorf("failed to read referenced secret %s%s for reference %s: %s", v1beta1constants.ReferencedResourcesPrefix, ref.ResourceRef.Name, resourceName, err)
```
vpnachev, line 725:
```suggestion
        return nil, fmt.Errorf("failed to read referenced secret %s%s for reference %s: %w", v1beta1constants.ReferencedResourcesPrefix, ref.ResourceRef.Name, resourceName, err)
```

It is better to use `%w` for errors, this could help up in the stack if the caller want to unwrap the error for some reason.

pkg/controller/lifecycle/actuator.go
```diff
@@ -695,6 +708,31 @@ func getShootResources(webhookCaBundle []byte, extensionNamespace, shootAccessSe
...
+    }
+
+    clientKeys, ok := refSecret.Data["keys"]
+    if ok != true {
```
vpnachev, line 729:
```suggestion
    if !ok {
```

pkg/controller/lifecycle/actuator.go
```diff
@@ -695,6 +708,31 @@ func getShootResources(webhookCaBundle []byte, extensionNamespace, shootAccessSe
...
+        return nil, fmt.Errorf("failed to extract public keys from secret. No data with name `keys` in secret.")
```
vpnachev, line 730: The error message can be simplified, it is enough just to mention that the secret is missing data key `keys` and this will be enough.

```suggestion
        return nil, fmt.Errorf("secret %s/%s is missing date key 'keys'", refSecret.Namespace, refSecret.Name)
```

docs/usage/shoot-extension.md
```diff
@@ -10,22 +10,59 @@ In most of the Gardener setups the `shoot-lakom-service` extension is enabled gl
...
 Lakom will validate all pods in all namespaces.
 
+### PublicKeysSecretReference
+
+Lakom, by default, tries to verify only workloads that belong to Gardener. Because of this, the only public keys that it uses to do its job are the ones for the Gardener workload.
+
+If you'd like to use Lakom as a tool for verifying your own workload, you'll need to add your own public keys to the ones that Lakom is already using. This can be achieved using Gardener [referenced resources](https://github.com/gardener/gardener/blob/master/docs/extensions/referenced-resources.md).
+
+Simply:
+1. Create a secret in your project namespace that contains a field `keys` with your keys as a value. Example keys:
```
vpnachev, line 48: It would be nice to document the format of a key (`name`, `algorithm` and `key` fields). Or for simplicity, add a reference to https://github.com/gardener/gardener-extension-shoot-lakom-service/blob/main/docs/usage/lakom.md#lakom-cosign-public-keys-configuration-file

pkg/controller/lifecycle/actuator.go
rrhubenov: Absolutely, good catch.

pkg/apis/lakom/types.go
```diff
@@ -34,4 +34,6 @@ type LakomConfig struct {
 
 	// The scope in which lakom will verify pods
 	Scope *ScopeType
+	// CosignPublicKeys is the cosign public keys used to verify image signatures.
+	PublicKeysSecretReference *string
```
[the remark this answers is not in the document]
rrhubenov: Sounds good to me, thank you for the detailed explanation

pkg/controller/lifecycle/actuator.go
```diff
@@ -153,18 +155,29 @@ func (a *actuator) Reconcile(ctx context.Context, logger logr.Logger, ex *extens
 		image.Tag = ptr.To[string](version.Get().GitVersion)
 	}
 
-	lakomConfig, err := yaml.JSONToYAML(a.serviceConfig.CosignPublicKeys.Raw)
+	gardenerPublicKeys, err := yaml.JSONToYAML(a.serviceConfig.CosignPublicKeys.Raw)
```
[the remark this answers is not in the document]
rrhubenov: My idea was to name them based on the source and purpose of the keys. One set of keys are coming from us, that being Gardener and validating our workload. The other set is coming from the shoot owner(client) and theoretically validating their workload. 

They are global in the sense that they get added to every instance of Lakom that we deploy, but they are not global in the sense of the workload that they are expected to validate.

So I personally would still prefer something that named them based on the target scope of images that they _should_ be able to validate.

pkg/admission/validator/lakom/shoot.go
```diff
@@ -73,9 +97,16 @@ func (s *shoot) Validate(_ context.Context, new, _ client.Object) error {
...
-	if lakomConfig.Scope == nil {
-		return nil
+
+	errList := field.ErrorList{}
+
+	if lakomConfig.Scope != nil {
+		errList = append(errList, s.validateScopeType(providerConfigPath.Child("scope"), *lakomConfig.Scope)...)
+	}
+	if lakomConfig.PublicKeysSecretReference != nil {
+		// TODO: Can the secret be checked?
```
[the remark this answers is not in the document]
rrhubenov: Could you elaborate on this?
You're saying that admission controller will have to fetch all secrets on the virtual cluster and find the one that corresponds to the resource?

pkg/controller/lifecycle/actuator.go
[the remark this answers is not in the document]
vpnachev: Ok, fine with me

pkg/admission/validator/lakom/shoot.go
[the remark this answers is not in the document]
rrhubenov: I see what you mean. But it turns out my question was simpler than that.
When reconciling the extension resource, we fetch the secret that is in the same namespace as the extension.

When validating the shoot on the runtime cluster, I thought there wasn't going to be a way to easily find the secret, since it's in the project namespace. Turns out the project namespace is the shoot namespace that we already have, so there's no issue here. It's just a simple get via the client.

I'll implement this the same way they do it in the registry cache: https://github.com/gardener/gardener-extension-registry-cache/blob/058fb0be0e353ab5c7a97a0b48be2c533da101e4/pkg/admission/validator/cache/shoot.go#L114-L135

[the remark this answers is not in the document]
rrhubenov: It was confusing to me that when reconciling the extension resource, we search for the secret from the shoot namespace on the seed, while when validating the shoot, we fetch the secret from the project namespace on the virtual cluster.

docs/usage/lakom.md
```diff
@@ -51,6 +51,11 @@ publicKeys:
     -----END PUBLIC KEY-----
 ```
 
+Here:
+- `name` can be set to any string.
+- `algorithm` needs to be chosen based on the key that you are adding.
```
vpnachev, line 56:
```suggestion
- `algorithm` is the algorithm that has to be used to verify the signature, see [Supported RSA Signature Verification Algorithms](#supported-rsa-signature-verification-algorithms) for the list of supported algorithms.
```

docs/usage/lakom.md
```diff
@@ -51,6 +51,11 @@ publicKeys:
...
+- `key` is the key that will be used for validating images using the algorithm that's specified.
```
vpnachev, line 57:
```suggestion
- `key` is the cryptographic public key that will be used for image signature validation.
```

pkg/apis/lakom/v1alpha1/types.go
```diff
@@ -19,4 +19,7 @@ type LakomConfig struct {
 	// The scope in which lakom will verify pods
 	// +optional
 	Scope *lakom.ScopeType `json:"scope"`
+	// CosignPublicKeys is the cosign public keys used to verify image signatures.
+	// +optional
+	TrustedKeysResourceName *string `json:"publicKeysSecretReference,omitempty"`
```
vpnachev, line 24:
```suggestion
	TrustedKeysResourceName *string `json:"trustedKeysResourceName,omitempty"`
```

pkg/apis/lakom/v1alpha1/types.go
```diff
@@ -19,4 +19,7 @@ type LakomConfig struct {
 	// The scope in which lakom will verify pods
 	// +optional
 	Scope *lakom.ScopeType `json:"scope"`
+	// CosignPublicKeys is the cosign public keys used to verify image signatures.
```
vpnachev, line 22:
```suggestion
	// TrustedKeysResourceName is the name of the shoot resource providing additional cosign public keys for image signature validation.
```

pkg/apis/lakom/types.go
```diff
@@ -34,4 +34,6 @@ type LakomConfig struct {
 
 	// The scope in which lakom will verify pods
 	Scope *ScopeType
+	// CosignPublicKeys is the cosign public keys used to verify image signatures.
+	TrustedKeysResourceName *string
```
vpnachev, line 38:
```suggestion
	// TrustedKeysResourceName is the name of the shoot resource providing additional cosign public keys for image signature validation.
	TrustedKeysResourceName *string
```

docs/usage/shoot-extension.md
```diff
@@ -10,22 +10,59 @@ In most of the Gardener setups the `shoot-lakom-service` extension is enabled gl
...
+      kind: Secret
+      name: lakom-secret
   extensions:
   - type: shoot-lakom-service
     disabled: true
     providerConfig:
       apiVersion: lakom.extensions.gardener.cloud/v1alpha1
       kind: LakomConfig
       scope: KubeSystem
+      publicKeysSecretReference: lakom-ref
```
vpnachev, line 26:
```suggestion
      trustedKeysResourceName: lakom-ref
```

docs/usage/shoot-extension.md
```diff
@@ -10,22 +10,59 @@ In most of the Gardener setups the `shoot-lakom-service` extension is enabled gl
...
+- name: example-client-key2
+  algorithm: RSASSA-PSS-SHA256
+  key: |-
+    -----BEGIN PUBLIC KEY-----
+    MFwwDQYJKoZIhvcNAQEBBQADSwAwSAJBAPeQXbIWMMXYV+9+j9b4jXTflnpfwn4E
+    GMrmqYVhm0sclXb2FPP5aV/NFH6SZdHDZcT8LCNsNgxzxV4N+UE/JIsCAwEAAQ==
+    -----END PUBLIC KEY-----
+```
+2. Add a reference to your secret via the `resources` field in the shoot spec as shown above.
+3. Add the name of your referenece in `publicKeysSecretReference` in the provider config as shown above.
```
vpnachev, line 66:
```suggestion
3. Add the name of your referenece in `trustedKeysResourceName ` in the provider config as shown above.
```

pkg/admission/validator/lakom/shoot.go
```diff
@@ -73,9 +141,16 @@ func (s *shoot) Validate(_ context.Context, new, _ client.Object) error {
...
-	if lakomConfig.Scope == nil {
-		return nil
+
+	allErrs := field.ErrorList{}
+
+	if lakomConfig.Scope != nil {
+		allErrs = append(allErrs, s.validateScopeType(providerConfigPath.Child("scope"), *lakomConfig.Scope)...)
+	}
+	if lakomConfig.TrustedKeysResourceName != nil {
+                allErrs = append(allErrs, s.validateTrustedKeys(ctx, providerConfigPath.Child("trustedKeysResourceName"), *lakomConfig.TrustedKeysResourceName, shoot.Spec.Resources, shoot.Namespace)...)
```
vpnachev, line 151: A bit hard to notice, but here you are using spaces instead of tabs for indentation.

pkg/admission/validator/lakom/shoot.go
```diff
@@ -73,9 +141,16 @@ func (s *shoot) Validate(_ context.Context, new, _ client.Object) error {
...
 	}
 
-	return s.validateScopeType(providerConfigPath.Child("scope"), *lakomConfig.Scope).ToAggregate()
+	return allErrs.ToAggregate()
+
```
vpnachev, line 155:
```suggestion
```

This empty line is not necessary.

pkg/admission/validator/lakom/shoot.go
```diff
@@ -44,16 +50,78 @@ func findExtension(extensions []core.Extension, extensionType string) (int, core
...
+	return errList
+}
+
+func (s *shoot) validateTrustedKeys(ctx context.Context, fldPath *field.Path, resourceName string, resources []core.NamedResourceReference, namespace string) field.ErrorList {
+    errList := field.ErrorList{}
+
+    ref := gardencorehelper.GetResourceByName(resources, resourceName)
+    if ref == nil {
+        return append(errList, field.Invalid(fldPath, resourceName, fmt.Sprintf("failed to find referenced resource with name %s", resourceName)));
```
vpnachev, line 90: There is no need to put the resource name in the `detail` parameter of the `field.Invalid` function as it is already set in the `value` parameter, otherwise the error message looks like 

```
Invalid value: "lakom-ref": failed to find referenced resource with name lakom-ref
```


You can use something like instead:
 ```suggestion
        return append(errList, field.Invalid(fldPath, resourceName, "there is no resource with this name in shoot.spec.resources"));
```

File: pkg/admission/validator/lakom/shoot.go [hunk 20]
```diff
@@ -44,16 +50,78 @@ func findExtension(extensions []core.Extension, extensionType string) (int, core
...
+    }
+    if ref.ResourceRef.Kind != "Secret" {
+        return append(errList, field.Invalid(fldPath, resourceName, fmt.Sprintf("resource with name %s is not of type secret", resourceName)));
```
vpnachev, line 93:
```suggestion
        return append(errList, field.Invalid(fldPath, resourceName, "resource must be of kind Secret"));
```

pkg/admission/validator/lakom/shoot.go
```diff
@@ -44,16 +50,78 @@ func findExtension(extensions []core.Extension, extensionType string) (int, core
...
+            ObjectMeta: metav1.ObjectMeta{
+                    Name:      ref.ResourceRef.Name,
+                    Namespace: namespace,
+            },
+    }
+
+    // Explicitly use the client.Reader to prevent controller-runtime to start Informer for Secrets
+    // under the hood. The latter increases the memory usage of the component.
+    if err := s.apiReader.Get(ctx, client.ObjectKeyFromObject(secret), secret); err != nil {
+            return append(errList, field.Invalid(fldPath, resourceName, fmt.Sprintf("failed to find secret %s from reference %s", client.ObjectKeyFromObject(secret), resourceName)))
```
vpnachev, line 106:
```suggestion
            return append(errList, field.Invalid(fldPath, resourceName, fmt.Sprintf("failed to get secret %s, %s", client.ObjectKeyFromObject(secret), err.Error())))
```

pkg/admission/validator/lakom/shoot.go
```diff
@@ -44,16 +50,78 @@ func findExtension(extensions []core.Extension, extensionType string) (int, core
...
+		} else if len(keys) != 1 {
+			errList = append(errList, field.Invalid(fldPath.Index(idx), k.Key, fmt.Sprintf("multiple keys with the name %s", k.Name)))
+		}
+	}
+
+	return errList
+}
+
+func (s *shoot) validateTrustedKeys(ctx context.Context, fldPath *field.Path, resourceName string, resources []core.NamedResourceReference, namespace string) field.ErrorList {
+    errList := field.ErrorList{}
```
vpnachev, line 86: The usage of `errList` is a bit odd, it always appended with single element and returned. 

Instead you can do 
```go
    return field.ErrorList{field.Invalid(...)}
```

pkg/admission/validator/lakom/shoot.go
```diff
@@ -44,16 +50,78 @@ func findExtension(extensions []core.Extension, extensionType string) (int, core
...
+    secret := &corev1.Secret{
+            ObjectMeta: metav1.ObjectMeta{
+                    Name:      ref.ResourceRef.Name,
+                    Namespace: namespace,
+            },
+    }
+
+    // Explicitly use the client.Reader to prevent controller-runtime to start Informer for Secrets
+    // under the hood. The latter increases the memory usage of the component.
+    if err := s.apiReader.Get(ctx, client.ObjectKeyFromObject(secret), secret); err != nil {
```
vpnachev, line 105: `client.ObjectKeyFromObject(secret)` is used multiple times, put into a variable and use the variable instead.

vpnachev (changes requested):

rrhubenov:
Thank you for the doc improvements @vpnachev 🙌
I've added a rough simple test for the `getClientKeys` function. I decided to test only it, since the new code in the `Reconcile` function is pretty simple and I don't see a need for it to be extracted to a new function.

I'll improve the test and add a couple new ones

pkg/lakom/config/config.go
```diff
@@ -87,9 +87,52 @@ func (c *Config) Complete() (*CompletedConfig, error) {
...
 	}
 
+	res.Keys = uniqueKeys(res.Keys)
+
 	return &res, nil
 }
 
+type PK interface {
+	Equal(x crypto.PublicKey) bool
+}
```
dimityrmirchev, line 97: This can be not exported or even declared in the function `uniqueKeys`.

pkg/lakom/config/config.go
```diff
@@ -87,9 +87,52 @@ func (c *Config) Complete() (*CompletedConfig, error) {
...
+		}
+	}
+
+	for _, idx := range idxUniq {
+		res = append(res, keys[idx])
+	}
+
+	return res
+}
+
```
dimityrmirchev, line 135: My version of this function 

```suggestion
// uniqueKeys returns a new VerifierKey slice that contains only the unique
// keys from the passed keys.
func uniqueKeys(keys []VerifierKey) []VerifierKey {
	// Equality between the keys is achieved via the 'Equal' function that we know
	// is implemented for every public key type. Ref: https://pkg.go.dev/crypto#PublicKey
	// Thus, casting all the keys to the pubKey interface enables us to call the function.
	type pubKey interface {
		Equal(x crypto.PublicKey) bool
	}

	var pubKeys []pubKey
	for _, k := range keys {
		pubKeys = append(pubKeys, k.Key.(pubKey))
	}

	var (
		res     []VerifierKey
		idxUniq []int
	)
	for idx, k1 := range pubKeys {
		uniq := true
		for _, uniqIdx := range idxUniq {
			k2 := pubKeys[uniqIdx]
			if k1.Equal(k2) {
				uniq = false
				break
			}
		}

		if uniq {
			idxUniq = append(idxUniq, idx)
		}
	}

	for _, idx := range idxUniq {
		res = append(res, keys[idx])
	}

	return res
}
```

dimityrmirchev (commented):
Two suggestions

pkg/lakom/config/config.go
dnaeon: Alternative and simpler version (IMO) of `uniqueKeys` would look like this.

``` go
// uniquekeys returns the unique set of [VerifierKey] items.
func uniqueKeys(keys []VerifierKey) []VerifierKey {
	result := make([]VerifierKey, 0)

	for _, item := range keys {
		key, ok := item.Key.(interface{ Equal(x crypto.PublicKey) bool })
		if !ok {
			continue
		}

		predicate := func(other VerifierKey) bool {
			return key.Equal(other.Key)
		}
		if !slices.ContainsFunc(result, predicate) {
			result = append(result, item)
		}
	}

	return result
}
```

pkg/lakom/config/config.go
```diff
@@ -87,9 +88,36 @@ func (c *Config) Complete() (*CompletedConfig, error) {
...
+	result := make([]VerifierKey, 0)
+
+	for _, item := range keys {
+		// Equality check between the keys is achieved via the 'Equal'
+		// function that we know is implemented for every public key type.
+		// Ref: https://pkg.go.dev/crypto#PublicKey
+		// We cast to the interface to inform the compiler of the func.
+		key, ok := item.Key.(interface{ Equal(x crypto.PublicKey) bool })
+		if !ok {
+			continue
```
rrhubenov, line 107: What should be done in this case? Would a simple log be enough?

pkg/lakom/config/config.go
rrhubenov: Thank you both for the improvements, I've used the approach of @dnaeon since I agree is the simplest :)

pkg/lakom/config/config.go
dnaeon: > This should theoretically happen only if the specific PublicKey has no implementation of the `Equal` function. That's possible to happen but I'd assume is extremely unlikely since I'd expect that the PublicKey key types that the Cosign lib supports should all be compliant with this interface.

The issue is that this interface in question is not actually defined by the stdlib and is more like a good practice, that all PublicKeys _should_ implement it.

I would keep this check, just to make sure that the PublicKeys I work with do implement it.

docs/usage/shoot-extension.md
```diff
@@ -10,22 +10,59 @@ In most of the Gardener setups the `shoot-lakom-service` extension is enabled gl
...
-Lakom will validate all pods in the `kube-system` namespace that are annotated with "managed-by/gardener"
+Lakom will validate all pods in the `kube-system` namespace that are annotated with "managed-by/gardener". This is the default value.
 - `Cluster`
 Lakom will validate all pods in all namespaces.
 
+### TrustedKeysResourceName
+
+Lakom, by default, tries to verify only workloads that belong to Gardener. Because of this, the only public keys that it uses to do its job are the ones for the Gardener workload.
+
+If you'd like to use Lakom as a tool for verifying your own workload, you'll need to add your own public keys to the ones that Lakom is already using. This can be achieved using Gardener [referenced resources](https://github.com/gardener/gardener/blob/master/docs/extensions/referenced-resources.md). More information about the keys and their format can be found [here](https://github.com/gardener/gardener-extension-shoot-lakom-service/blob/main/docs/usage/lakom.md#lakom-cosign-public-keys-configuration-file).
```
vpnachev, line 45: You can keep the references to other files in the repo local

```suggestion
If you'd like to use Lakom as a tool for verifying your own workload, you'll need to add your own public keys to the ones that Lakom is already using. This can be achieved using Gardener [referenced resources](https://github.com/gardener/gardener/blob/master/docs/extensions/referenced-resources.md). More information about the keys and their format can be found [here](lakom.md#lakom-cosign-public-keys-configuration-file).
```

pkg/admission/validator/lakom/shoot.go
```diff
@@ -44,16 +50,78 @@ func findExtension(extensions []core.Extension, extensionType string) (int, core
...
 	errList := field.ErrorList{}
-
 	if !lakom.AllowedScopes.Has(scopeType) {
 		errList = append(errList, field.NotSupported(fldPath, scopeType, lakom.AllowedScopes.UnsortedList()))
 	}
 
 	return errList
 }
 
+// TODO: This check exists in the validation of the lakom config as well. It can be extracted as a util function
```
vpnachev, line 60: Will you address this TODO with this PR or not?

pkg/admission/validator/lakom/shoot.go
```diff
@@ -44,16 +50,78 @@ func findExtension(extensions []core.Extension, extensionType string) (int, core
...
+	usedNames := map[string]any{}
+	for idx, k := range cosignPublicKeys {
+		if k.Name == "" {
+			errList = append(errList, field.Required(fldPath.Index(idx), "key name should no be empty"))
+		}
+
+		if _, ok := usedNames[k.Name]; ok {
+			errList = append(errList, field.Duplicate(fldPath.Index(idx), k.Name))
+		}
+		usedNames[k.Name] = nil
```
vpnachev, line 73: I think this code should be skipped when the name is empty, what do you think?

pkg/controller/lifecycle/actuator.go
```diff
@@ -153,18 +155,30 @@ func (a *actuator) Reconcile(ctx context.Context, logger logr.Logger, ex *extens
...
+	if lakomProviderConfig.TrustedKeysResourceName != nil {
+		var err error
+		clientPublicKeys, err = getClientKeys(ctx, a.client, cluster.Shoot.Spec.Resources, *lakomProviderConfig.TrustedKeysResourceName, namespace)
+		if err != nil {
+			return fmt.Errorf("failed to get the additional keys: %w", err)
+		}
+	}
+
+	lakomPublicKeys := gardenerPublicKeys
+	lakomPublicKeys = append(lakomPublicKeys, clientPublicKeys...)
```
vpnachev, line 173:
```suggestion
	lakomPublicKeys := append(gardenerPublicKeys, clientPublicKeys...)
```

pkg/controller/lifecycle/actuator_test.go
```diff
@@ -164,6 +170,120 @@ var _ = Describe("Actuator", func() {
...
+
+		It("Should return an err if the resource is not found", func() {
+			_, err := getClientKeys(ctx, fakeclient, resources[0:0], resourceName, namespace)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("failed to find referenced resource with name " + resourceName))
+		})
+
+		It("Should return an err if the reference is found but the resource with the given name is not found", func() {
+			resourcesWrongName := resources
+			wrongName := "non-existant"
```
vpnachev, line 262:
```suggestion
			wrongName := "non-existent"
```

pkg/controller/lifecycle/actuator_test.go
```diff
@@ -164,6 +170,120 @@ var _ = Describe("Actuator", func() {
...
+		})
+
+		It("Should return an err if the resource is not found", func() {
+			_, err := getClientKeys(ctx, fakeclient, resources[0:0], resourceName, namespace)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("failed to find referenced resource with name " + resourceName))
+		})
+
+		It("Should return an err if the reference is found but the resource with the given name is not found", func() {
+			resourcesWrongName := resources
```
vpnachev, line 261: This is not making a deep copy of the slice and the changes in `resourcesWrongName` are reflected also in `resources`.

pkg/lakom/config/config.go
```diff
@@ -87,9 +88,36 @@ func (c *Config) Complete() (*CompletedConfig, error) {
 		res.Keys = append(res.Keys, config)
 	}
 
+	res.Keys = uniqueKeys(res.Keys)
+
 	return &res, nil
 }
 
+// uniquekeys returns the unique set of [VerifierKey] items.
```
vpnachev, line 96:
```suggestion
// uniqueKeys returns the unique set of [VerifierKey] items.
```

pkg/lakom/config/config_suite_test.go
```diff
@@ -0,0 +1,17 @@
...
+package config
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestCMD(t *testing.T) {
```
vpnachev, line 14: I see where this is coming from, but `CMD` is not best name here. Usually the naming scheme is `Test<package-name>`.

 ```suggestion
func TestConfig(t *testing.T) {
```

File: pkg/lakom/config/config.go (hunk 26 above)
Comment (vpnachev) [reply pkg/lakom/config/config.go:107]: I think an error should be returned here. Or at least check in the `Complete` method that the keys do implement the interface. Another option is to change the type of `Verifier.Key` to the interface that `crypto.PublicKey` was supposed to be but it is not due to backward compatibility. In this case, you can be sure in `uniqueKeys` that the key implements the interface.

File: pkg/lakom/config/config_suite_test.go (hunk 34 above)
Comment (rrhubenov) [reply pkg/lakom/config/config_suite_test.go:14]: I actually missed renaming this. So I didn't intend for it to be called CMD. Thanks

File: pkg/controller/lifecycle/actuator.go (hunk 30 above)
Comment (rrhubenov) [reply pkg/controller/lifecycle/actuator.go:173]: This was actually the initial implementation. But one of the go linters complained about assigning the result of append to a different slice. Ref https://concourse.ci.gardener.cloud/builds/357267772

I'm not fond of the way it's now, but this was the easiest/cleanest alternative I could think of

File: pkg/controller/lifecycle/actuator_test.go (hunk 32 above)
Comment (rrhubenov) [reply pkg/controller/lifecycle/actuator_test.go:261]: Yeah, silly mistake. Good catch

File: pkg/lakom/config/config.go (hunk 26 above)
Comment (rrhubenov) [reply pkg/lakom/config/config.go:107]: > I think an error should be returned here.
 
In the case that a user supplies a public key type that has no implementation of `Equal`, and we throw an error, we'd effectively be disallowing them from using Lakom with the specific key that has the issue and they want to use. I'm worried that that might cause more issues in the future rather than make things better.

I personally think we should output some sort of warning. Though that would mean that duplicate keys would be possible in this case.

> Another option is to change the type of Verifier.Key to the interface that crypto.PublicKey was supposed to be but it is not due to backward compatibility. In this case, you can be sure in uniqueKeys that the key implements the interface.

Won't this just move the check earlier into the codepath? 
The thing I like with the current approach is that the topic of 'comparing keys' is contained completely inside this function. So the whole issue does not leak to any other types/functions/files and you don't need to know about these things unless you're trying to modify/use the `uniqueKeys` function.

File: pkg/admission/validator/lakom/shoot.go (hunk 28 above)
Comment (rrhubenov) [reply pkg/admission/validator/lakom/shoot.go:60]: So the original code that does these checks works a bit differently with the way it handles the errors:
https://github.com/gardener/gardener-extension-shoot-lakom-service/blob/9ee7121684091d71148300c087a2341c929cb111/pkg/lakom/config/config.go#L42-L57

I would like to extract the common logic but I'm not sure how to reconcile this difference between them.
Maybe the new function could also return some sort of aggregate of all the errors and return it? What do you think?

File: pkg/admission/validator/lakom/shoot.go (hunk 29 above)
Comment (rrhubenov) [reply pkg/admission/validator/lakom/shoot.go:73]: Yeah, absolutely agree

File: pkg/admission/validator/lakom/shoot.go (hunk 28 above)
Comment (vpnachev) [reply pkg/admission/validator/lakom/shoot.go:60]: The common function can return `field.ErrorList`, then
- `github.com/gardener/gardener-extension-shoot-lakom-service/pkg/admission/validator/lakom.validateCosignPublicKeys` will return directly the result returned from the common function
- `github.com/gardener/gardener-extension-shoot-lakom-service/pkg/lakom/config.validate` will call `ToAggregate()` on the result returned from the common function and return it.

File: pkg/lakom/config/config.go (hunk 26 above)
Comment (vpnachev) [reply pkg/lakom/config/config.go:107]: Until now it was not allowed end users to provide trusted keys, hence we can still implement restrictions what keys are supported without breaking any existing usage (there is still nobody using the feature :wink:)

The validation of the correctness of the provided keys should not be implemented only in the `lakom` component, it must be implemented also in the `shoot-lakom-admission` so that end user are not allowed to set key that does not implement the public key interface.

File: pkg/admission/validator/lakom/shoot.go (hunk 28 above)
Comment (rrhubenov) [reply pkg/admission/validator/lakom/shoot.go:60]: Me and @vpnachev agreed to leave it as is. Resolving

File: pkg/controller/lifecycle/actuator.go (hunk 30 above)
Comment (rrhubenov) [reply pkg/controller/lifecycle/actuator.go:173]: I thought about renaming gardenerPublicKeys to lakomPublicKeys, but that conflicts with my wishes here https://github.com/gardener/gardener-extension-shoot-lakom-service/pull/112#discussion_r1810375531 😃
So I'd rather we leave it as it is

File: pkg/lakom/config/config.go [hunk 35]
```diff
@@ -87,9 +88,36 @@ func (c *Config) Complete() (*CompletedConfig, error) {
...
+		// Ref: https://pkg.go.dev/crypto#PublicKey
+		// We cast to the interface to inform the compiler of the func.
+		key, ok := item.Key.(interface{ Equal(x crypto.PublicKey) bool })
+		if !ok {
+			continue
+		}
+
+		predicate := func(other VerifierKey) bool {
+			return key.Equal(other.Key)
+		}
```
vpnachev, line 112: It would be nice to define this predicate only once, i.e. outside the for loop.

vpnachev (approved):
Just one suggestion, rest is looking good to me.

pkg/lakom/config/config.go
rrhubenov: Started moving the predicate outside the loop but realised the function actually depends on the `key` var that gets defined in the loop. So this requires that the predicate gets redefined on each loop.

vpnachev: Good catch, sorry for the confusion.

PR #114, opened 2024-10-18
Fix lakom permissions when extension is using scope=Cluster

vpnachev:
Fix lakom permissions when extension is using scope=Cluster

Fixes #

**Release note**:

```bugfix operator
Fix permissions of lakom admission controller when it is installed with `Cluster` scope for a shoot cluster.
```


dimityrmirchev (approved):
/lgtm

PR #126, opened 2025-01-09
Verify Images/Charts for Pods, ControllerDeployments, Gardenlets & operator.Extensions

### LINKED ISSUE #97: Verify helm charts stored in OCI registry
Helm charts stored in OCI registries to be verified for signatures, just like OCI images. 
Gardener is planning to use remote helm charts in the following APIs as of now:
- [controllerdeployments.core.gardener.cloud](https://github.com/gardener/gardener/blob/c2418171a3d06837e839f22217bc0cb6a21f3764/pkg/apis/core/v1/types_controllerdeployment.go#L48-L50)
- [gardenlets.seedmanagement.gardener.cloud](https://github.com/gardener/gardener/blob/c2418171a3d06837e839f22217bc0cb6a21f3764/pkg/apis/seedmanagement/v1alpha1/types_gardenlet.go#L73-L74)
- [extensions.operator.gardener.cloud](https://github.com/gardener/gardener/blob/c2418171a3d06837e839f22217bc0cb6a21f3764/pkg/apis/operator/v1alpha1/types_extension.go#L111-L113)

**Why is this needed**:
To improve security of the supply chain by ensuring only trusted artifacts are deployed.

rrhubenov:
Handling of the admission request now passes the request through a layer that generates a list of generic targets for verification.

This way, we can reuse the existing verification logic.

Verification of helm charts is required. 

Fixes #97 

**Release note**:

```other operator

```

### DIFF
--- a/pkg/lakom/resolvetag/admission.go
+++ b/pkg/lakom/resolvetag/admission.go
@@ -164,7 +164,11 @@ func (h *handler) Handle(ctx context.Context, request admission.Request) admissi
 func (h *handler) handlePod(ctx context.Context, p *corev1.Pod, logger logr.Logger) error {
 	logger.Info("Handling new pod request")
 
-	kcr := utils.NewLazyKeyChainReaderFromPod(ctx, h.reader, p, h.useOnlyImagePullSecrets)
+	var secretNames []string
+	for _, ips := range p.Spec.ImagePullSecrets {
+		secretNames = append(secretNames, ips.Name)
+	}
+	kcr := utils.NewLazyKeyChainReaderFromPod(ctx, h.reader, p.Namespace, secretNames, h.useOnlyImagePullSecrets)
 
 	for idx, ic := range p.Spec.InitContainers {
 		image, err := h.handleContainer(ctx, ic.Image, kcr, logger.WithValues("initContainer", ic.Name))
--- a/pkg/lakom/utils/utils.go
+++ b/pkg/lakom/utils/utils.go
@@ -42,13 +42,13 @@ func NewLazyKeyChainReader(reader func() (authn.Keychain, error)) *lazyKeyChainR
 }
 
 // NewLazyKeyChainReaderFromPod creates lazyKeyChainReader for given pod.
-func NewLazyKeyChainReaderFromPod(ctx context.Context, c client.Reader, pod *corev1.Pod, useOnlyImagePullSecrets bool) *lazyKeyChainReader {
+func NewLazyKeyChainReaderFromPod(ctx context.Context, c client.Reader, namespace string, secretNames []string, useOnlyImagePullSecrets bool) *lazyKeyChainReader {
 	return NewLazyKeyChainReader(
 		func() (authn.Keychain, error) {
-			var imagePullSecrets = make([]corev1.Secret, len(pod.Spec.ImagePullSecrets))
-			for _, sn := range pod.Spec.ImagePullSecrets {
+			var imagePullSecrets = make([]corev1.Secret, len(secretNames))
+			for _, sn := range secretNames {
 				secret := &corev1.Secret{}
-				secretKey := client.ObjectKey{Namespace: pod.GetNamespace(), Name: sn.Name}
+				secretKey := client.ObjectKey{Namespace: namespace, Name: sn}
 
 				if err := c.Get(ctx, secretKey, secret); err != nil {
 					if apierrors.IsNotFound(err) {
--- a/pkg/lakom/verifysignature/admission.go
+++ b/pkg/lakom/verifysignature/admission.go
@@ -7,14 +7,17 @@ package verifysignature
 import (
 	"context"
 	"fmt"
-	"net/http"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/gardener/gardener-extension-shoot-lakom-service/pkg/lakom/config"
 	"github.com/gardener/gardener-extension-shoot-lakom-service/pkg/lakom/metrics"
 	"github.com/gardener/gardener-extension-shoot-lakom-service/pkg/lakom/utils"
 
+	"github.com/gardener/gardener/pkg/apis/core"
+	operatorv1alpha1 "github.com/gardener/gardener/pkg/apis/operator/v1alpha1"
+	"github.com/gardener/gardener/pkg/apis/seedmanagement"
 	"github.com/go-logr/logr"
 	admissionv1 "k8s.io/api/admission/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -51,13 +54,13 @@ func (hb HandleBuilder) WithManager(mgr manager.Manager) HandleBuilder {
 	return hb
 }
 
-// WithUseOnlyImagePullSecrets sets only the image pull secrets to be used to access the OCI Registry.
+// WithUseOnlyImagePullSecrets sets only the artifact pull secrets to be used to access the OCI Registry.
 func (hb HandleBuilder) WithUseOnlyImagePullSecrets(useOnlyImagePullSecrets bool) HandleBuilder {
 	hb.useOnlyImagePullSecrets = useOnlyImagePullSecrets
 	return hb
 }
 
-// WithAllowUntrustedImages configures the webhook to allow images without trusted signature.
+// WithAllowUntrustedImages configures the webhook to allow artifacts without trusted signature.
 func (hb HandleBuilder) WithAllowUntrustedImages(allowUntrustedImages bool) HandleBuilder {
 	hb.allowUntrustedImages = allowUntrustedImages
 	return hb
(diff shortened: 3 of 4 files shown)


rrhubenov:
@vpnachev Hi Vladi, I'd like your opinion on the changes till now. After we come to an agreement for the final version of this change, I'd like to wait for us to merge https://github.com/gardener/gardener-extension-shoot-lakom-service/pull/131 before I add tests.

pkg/lakom/utils/utils.go
```diff
@@ -67,3 +67,30 @@ func NewLazyKeyChainReaderFromPod(ctx context.Context, c client.Reader, pod *cor
...
+					return nil, err
+				}
+				imagePullSecrets = append(imagePullSecrets, *secret)
+			}
+
+			if useOnlyImagePullSecrets {
+				return kauth.NewFromPullSecrets(ctx, imagePullSecrets)
+			}
+
+			return k8schain.NewFromPullSecrets(ctx, imagePullSecrets)
```
vpnachev, line 93: It would be nice to re-use this code in `NewLazyKeyChainReaderFromPod`

pkg/lakom/resolvetag/admission.go
```diff
@@ -128,8 +136,8 @@ func (h *handler) Handle(ctx context.Context, request admission.Request) admissi
 	ctx, cancel := context.WithTimeout(ctx, time.Second*25)
 	defer cancel()
 
-	if request.Kind != podGVK {
-		return admission.Allowed("resource is not v1.Pod")
+	if !allowedResources.Has(request.Kind) {
```
vpnachev, line 139: I think it is better to use switch/case here by the resource GVK and call the specific implementation. 
For each supported GVK there should be a handler method resolving the tag to digest

```go
	var (
		patch []byte
		err   error
	)

	switch request.Kind {
	case podGVK:
		patch, err = h.handlePod(ctx, request)
	case controllerDeploymentGVK:
		patch, err = h.handleControllerDeployment(ctx, request)
	case gardenletGVK:
		patch, err = h.handleGardenlet(ctx, request)
	case extensionGVK:
		patch, err = h.handleExtension(ctx, request)
	default:
		return admission.Allowed(fmt.Sprintf("resource is not one of %v", allowedResources.UnsortedList()))
	}

	if err != nil {
		return admission.Errored(http.StatusInternalServerError, err)
	}

	return admission.PatchResponseFromRaw(request.Object.Raw, patch)
```

The body of each `handle<GVK>` method is already implemented in the `handleObject`, using dedicated methods will improve code readability, also it should be easier to test `handle<GVK>` instead of the generic `handleObject` method.

For example `handleGardenlet` should look like this
```go 
func (h *handler) handleGardenlet(ctx context.Context, request admission.Request) ([]byte, error) {
	var (
		logger    = h.logger.WithValues("controller-deployment", client.ObjectKey{Name: request.Name})
		kcr       = utils.NewLazyKeyChainReaderFromSecrets(ctx, h.reader, request.Namespace, []string{}, h.useOnlyImagePullSecrets)
		gardenlet = seedmanagement.Gardenlet{}
	)

	if err := h.decoder.Decode(request, &gardenlet); err != nil {
		return nil, err
	}

	resolved, err := h.handleContainer(ctx, gardenlet.Spec.Deployment.Helm.OCIRepository.GetURL(), kcr, logger)
	if err != nil {
		return nil, err
	}
	gardenlet.Spec.Deployment.Helm.OCIRepository.Ref = &resolved

	if gardenlet.Spec.Deployment.Image != nil {
		resolved, err = h.handleContainer(ctx, getURL(gardenlet.Spec.Deployment.Image), kcr, logger)
		if err != nil {
			return nil, err
		}
		// TODO(rado): Hmmm, I'm not sure this should be the change.
		gardenlet.Spec.Deployment.Image.Tag = &resolved
	}

	return json.Marshal(gardenlet)
}
```

pkg/lakom/resolvetag/admission.go
```diff
@@ -140,63 +148,169 @@ func (h *handler) Handle(ctx context.Context, request admission.Request) admissi
...
+
+	if img.Tag != nil {
+		ref = ref + ":" + *img.Tag
 	}
 
-	return nil
+	return strings.TrimPrefix(ref, "oci://")
 }
 
 func (h *handler) handleContainer(ctx context.Context, image string, kcr utils.KeyChainReader, logger logr.Logger) (string, error) {
```
vpnachev, line 316: Now it makes sense to rename this method to `handleImage` or `resolveImage` because it is no longer working on container images only.

pkg/lakom/resolvetag/admission.go
rrhubenov: I agree, code readability will improve this way. The easier tests will be a big plus. Thanks!

pkg/lakom/resolvetag/admission.go
```diff
@@ -114,8 +118,12 @@ type handler struct {
 }
 
 var (
-	podGVK               = metav1.GroupVersionKind{Group: "", Kind: "Pod", Version: "v1"}
-	controlledOperations = sets.NewString(string(admissionv1.Create), string(admissionv1.Update))
+	podGVK                  = metav1.GroupVersionKind{Group: "", Kind: "Pod", Version: "v1"}
+	controllerDeploymentGVK = metav1.GroupVersionKind{Group: "core.gardener.cloud", Kind: "ControllerDeployment", Version: "v1"}
```
vpnachev, line 122: Will this GVK also catch the `v1beta1` version of the ControllerDeployment API?
Not sure, but maybe it is better to use `metav1.GroupKind` in order to make it work for multiple versions. 

PS. We should take care to configure `matchPolicy` and `admissionReviewVersions` in the `(Mutating|Validating)WebhookConfigurations`

pkg/lakom/resolvetag/admission.go
```diff
@@ -125,81 +133,195 @@ func (h *handler) GetLogger() logr.Logger {
...
+	switch request.Kind {
+	case podGVK:
+		pod := corev1.Pod{}
+		err = h.decoder.Decode(request, &pod)
+		if err != nil {
+			break
+		}
+		patch, err = h.handlePod(ctx, pod)
+	case controllerDeploymentGVK:
+		controllerDeployment := core.ControllerDeployment{}
```
vpnachev, line 162: Why is the internal API version used here? Clients should use versioned APIs only, ref: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#:~:text=Clients%20should%20consume%20and%20operate%20on%20the%20versioned%20APIs%20exclusively.

rrhubenov: Sorry, I hope you bear with me since I might be confused on this topic.
Wouldn't we want to convert the external version (corev1, v1alpha1, etc.) into the internal version,
then do whatever we do with it, and finally convert it back to an external version before sending it back
"through the wire". 
Meaning that my mistake here is that I'm decoding the `Pod` and the `Extension` into the external version, 
rather than the internal one.

If I convert to an external version, I'll have to handle all the different versions explicitly. 
The example given in the spec from the doc you linked goes through the process I outlined.

pkg/lakom/resolvetag/admission.go
rrhubenov: As far as I can tell, the `admissionReviewVersions` field relates to the version of the `AdmissionReview` object that represents the request. Would we need to change it?

Regarding the `matchPolicy` -- thank you for bringing it to my attention. I hadn't looked deep enough into what it did and how it worked. I'll update it.

pkg/lakom/resolvetag/admission.go
rrhubenov: Now that I took a look into the `matchPolicy` field In the `(Mutating|Admission)WebhookConfiguration` objects, I can see that we can always expect the version we've asked for in the `rules` when the `matchPolicy` has a value of `Equivalent`. So we'll have to handle only that version.

My understanding now is that we'd want webhooks to always use the external versions, rather than the internal one. Since that's the actual object we're modifying/validating. But again, I'm still a bit confused why we wouldn't want to convert into the internal representation of the resource. 
The kube-apiserver already does a transformation into the internal version and then back to the external version
we've requested from the rule (when the `matchPolicy` is `Equivalent`).

charts/lakom/templates/mutatingwebhookconfiguration.yaml
```diff
@@ -46,6 +46,9 @@ webhooks:
     resources:
     - pods
     - pods/ephemeralcontainers
+    - controllerdeployments
+    - gardenlets
+    - operator.extensions
```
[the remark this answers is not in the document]
rrhubenov: As we discussed, I've extracted the rules for the web hooks into the values.yaml file for the lakom chart.
This will allow users to choose which of the resources need to be deployed.
I've also added some documentation on the topic

pkg/lakom/resolvetag/admission.go
```diff
@@ -125,81 +133,195 @@ func (h *handler) GetLogger() logr.Logger {
...
+
+	return json.Marshal(pod)
+}
+
+// getURL returns the fully-qualified OCIRepository URL of the image.
+func getURL(img *seedmanagementv1alpha1.Image) string {
+	ref := *img.Repository
+
+	if img.Tag != nil {
+		ref = ref + ":" + *img.Tag
```
vpnachev, line 318: I guess the tag can contain sha256 digest, if that is the case - the ref should be 
```go
ref + "@" + *img.Tag
```

docs/usage/lakom.md
```diff
@@ -64,3 +64,22 @@ Here:
...
+- `controllerdeployments.core.gardener.cloud/v1`
+- `gardenlets.seedmanagement.gardener.cloud/v1alpha1`
+- `extensions.operator.gardener.cloud/v1alpha1`
+
+```
+When deploying Lakom via the helm chart in `/charts/lakom`, the `rules` key
+can be fully customized to include any of the listed resources above.
+Make sure that they are registered with the same group & versions as the ones listed above.
+Any difference will cause Lakom to fail the decoding of the resources and
+automatically reject them.
```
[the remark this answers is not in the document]
rrhubenov: Whoops, you're right

docs/usage/lakom.md
```diff
@@ -64,3 +64,22 @@ Here:
...
+```
+
+
+
```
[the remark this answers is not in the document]
rrhubenov: Thanks, this is exactly what I wanted but wasn't sure what would be best.

pkg/lakom/verifysignature/admission.go
```diff
@@ -135,106 +139,262 @@ type handler struct {
...
+
+	return verificationTargets, kcr, nil
+}
+
+// getURL returns the fully-qualified OCIRepository URL of the image.
+func getURL(img *seedmanagementv1alpha1.Image) string {
+	ref := *img.Repository
+
+	if img.Tag != nil {
+		ref = ref + ":" + *img.Tag
```
vpnachev, line 370: Same as https://github.com/gardener/gardener-extension-shoot-lakom-service/pull/126#discussion_r1949963870, the tag could be a digest.

vpnachev (commented):
Just one last thing, rest looks good

vpnachev (approved):
/lgtm

PR #131, opened 2025-01-27
Fake registry for tests

### LINKED ISSUE #37: Rework unit tests to not rely on external resources
Currently, some unit tests interact with external OCI registries like GCR and docker hub. Examples:

https://github.com/gardener/gardener-extension-shoot-lakom-service/blob/d60a2638f2c3b6d2381b745a614f663039e4359c/pkg/lakom/resolvetag/resolver_test.go#L51-L79

https://github.com/gardener/gardener-extension-shoot-lakom-service/blob/d60a2638f2c3b6d2381b745a614f663039e4359c/pkg/lakom/verifysignature/verifier_test.go#L56-L84

It would be better the OCI registries to be mocked.

**Why is this needed**:
The unit tests will fail in case of network issue, OCI registry downtime, server side rate limits, or simply deleted remote resources.

### LINKED ISSUE #128: Tests using a fake registry
Non E2E tests should not rely on a real deployment of an artifact registry.
In those cases, we can leverage some sort of fake registry implementation that can be passed around the verification/resolving functions.
**Why is this needed**:
More reliable tests that do not depend on the network or a publicly/privately deployed registry outside the local network.

One possibility is to setup tests the way [go-containeregistry](https://github.com/google/go-containerregistry) does.
They have a pkg called [registry](https://github.com/google/go-containerregistry/tree/main/pkg/registry) that "implements a Docker v2 registry and the OCI distribution specification ... It is designed to be used anywhere a low dependency container registry is needed, with an initial focus on tests."

Example usage in tests can be found [here](https://github.com/google/go-containerregistry/blob/6bce25ecf0297c1aa9072bc665b5cf58d53e1c54/pkg/v1/remote/multi_write_test.go#L107-L114).

rrhubenov:
Faster, more stable tests. Easier to perform them, since we can now upload the objects we want to verify, rather than
relying on ones already existing in public registries.

Should potentially make the tests more robust.
Fixes #128 
Fixes #37 

Setup code has been moved to a BeforeSuite ginkgo function that has been put in the `*_suite_test.go` files. 
There is most likely a way with which the code for creating and pushing the signature for an image can be improved, but I found the sigstore libraries very hard to work with and could not implement something simpler.

**Release note**:

```other developer

```

### DIFF
--- a/docs/oci_registries.md
+++ b/docs/oci_registries.md
@@ -113,3 +113,46 @@ then verifying them. In theory, we wouldn't need to use Cosign for this, but
 it uses a popular format for the signatures and the libraries help us 
 save work.
 
+In the context of Lakom, what interests us is how Cosign signs OCI artifacts
+and how it stores them.
+The signing process in a nutshell is the following -- `Sign(sha256(SimpleSigningPayload(sha256(Image Manifest))))`, where:
+- Image Manifest is what it says. Just the manifest of the image (artifact).
+- sha256 is the hashing algorithm used.
+- SimpleSigningPayload is the interesting part here. Instead of just signing the
+hash of the manifest itself, the hash itself is packed into a JSON doc that
+is called the `SimpleSigningPayload`. Example payload:
+```json
+{
+    "critical": {
+           "identity": {
+               "docker-reference": "testing/manifest"
+           },
+           "image": {
+               "Docker-manifest-digest": "sha256:20be...fe55"
+           },
+           "type": "cosign container image signature"
+    },
+    "optional": {
+           "creator": "atomic",
+           "timestamp": 1458239713
+    }
+}
+```
+
+We can see that the hash of the manifest is stored in the `Docker-manifest-digest`.
+Now this object is the one that gets signed using our private key. The signed
+version of this objest is the one that gets stored and gets used for
+verification. 
+
+Storage of the signature is done via an image manifest that gets uploaded
+in the same repo as the image, but tagged with `<image-digest>.sig`.
+Thus, the signature discovery mechanism is to just fetch the image digest 
+that is tagged with the aforementioned tag.
+
+More info on this whole process can be found [here](https://github.com/sigstore/cosign/blob/main/specs/SIGNATURE_SPEC.md).
+
+Lakom relies on the cosign libraries to automate this whole process. When
+validating an artifact, if we know its digest, we can fetch the signature and
+verify it. If the signature is valid, we can be sure that the artifact has not
+been tampered with.
+
--- a/docs/oci_registries.md
+++ b/docs/oci_registries.md
@@ -1,4 +1,4 @@
-# Crash Course on OCI Registries
+# OCI Registries
 
 This document aims to introduce the reader to the general architecture of OCI
 Registries, their history and jargon that might be encountered.
--- a/docs/oci_registries.md
+++ b/docs/oci_registries.md
@@ -145,7 +145,7 @@ version of this objest is the one that gets stored and gets used for
 verification. 
 
 Storage of the signature is done via an image manifest that gets uploaded
-in the same repo as the image, but tagged with `<image-digest>.sig`.
+in the same repo as the image, but tagged with `sha256-<container-image-digest>.sig`.
 Thus, the signature discovery mechanism is to just fetch the image digest 
 that is tagged with the aforementioned tag.
 
--- a/docs/oci_registries_crash_course.md
+++ b/docs/oci_registries_crash_course.md
@@ -55,15 +55,15 @@ it can be stored only once and referenced by multiple images.
 
 ### Manifests
 
-A manifest is just a JSON document. It is stored seperately from the blobs.
+A manifest is just a JSON document. It is stored separately from the blobs.
 But similarly to the blobs, it is also content addressable. The manifest gets
-formated to a canonical form and then hashed. The hash becomes the address of
+formatted to a canonical form and then hashed. The hash becomes the address of
 the manifest.
 
 Where manifests become powerful is when they are used to reference to other
 manifests or blobs. These references are called [OCI Content Descriptors](https://github.com/opencontainers/image-spec/blob/v1.0.1/descriptor.md).
-For example, the Image Manifest that most readers would be familiar with containes
-a lisf of layers that are used to build the image. Each layer is a reference to
+For example, the Image Manifest that most readers would be familiar with contains
+a list of layers that are used to build the image. Each layer is a reference to
 a blob.
 
 An additional reference outside the layer references is used for storing 
@@ -71,7 +71,7 @@ additional configuration. More info can be found in the official [OCI Image Mani
 
 The term _artifact_ is just a generalization over the Image Manifest idea.
 The Image Manifest has a specific format that can be found in the aforementioned [OCI Image Manifest spec](https://github.com/opencontainers/image-spec/blob/v1.0.1/manifest.md).
-The important field is the .config.mediaType field. Based on the official guidelines,
+The important field is the `.config.mediaType` field. Based on the official guidelines,
 it can be used for defining types other than an image. More info can be found [here](https://github.com/opencontainers/image-spec/blob/main/manifest.md#guidelines-for-artifact-usage).
 
 Combining all of these ideas, we can see that artifacts can be represented by 1 
@@ -93,7 +93,7 @@ to address them in a human-readable format. Tags allow us to add a reference
 to a given manifest that differs from the digest of the object.
 
 Normally, whenever we push an image to a registry, if we don't give it an
-explicit tag, the registry would most likey give it the tag `latest`. This,
+explicit tag, the registry would most likely give it the tag `latest`. This,
 actually is not mandatory based on the distribution spec. Manifests don't 
 require that they have a tag.
 
--- a/docs/oci_registries_crash_course.md
+++ b/docs/oci_registries_crash_course.md
@@ -1,0 +1,115 @@
+# Crash Course on OCI Registries
+
+This document aims to introduce the reader to the general architecture of OCI
+Registries, their history and jargon that might be encountered.
+
+## History of Container Registries
+
+Most readers might be familiar with the container registry that was introduced
+by Docker. It aims to provide a way for developers to store and distribute
+container images that can be used for running applications in containers.
+
+In the beginning, the registry was meant to only host images in a way that the
+different layers of the image could be downloaded in parallel, while being 
+assembled on the client side.
+
+In June 2015, multiple companies agreed to standardize the container image format,
+including its distribution and runtime. This was the birth of the Open Container
+Initiative (OCI). 
+
+
+## Understanding OCI Registries
+
+The most important document that pertains to OCI Registries in the case of Lakom is the [OCI 
+distribution spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md). 
+It outlines the way that artifacts shall be stored and fetched from a registry 
+that adheres to the OCI standard. 
+
+OCI Registries are meant to store _content_. While in the beginning the main
+problem that they were trying to solve was image storage, during the evolution
+of the image registry concept, a decision was made to make it more general
+for storage of any form of objects, sometimes called _artifacts_.
+
+An artifact is an abstract idea of multiple components that together describe
+how an object is built along with storing metadata about the object.
+
+To implement this, every OCI registry has to expose an API for creating
+3 types of objects:
+- Blobs
+- Manifests
+- Tags
+
+
+### Blobs
+
+A blob is just a set of bytes. Nothing more. The important part about blobs
+is that they are _content addressable_. [Content Addressable Storage (CAS)](https://en.wikipedia.org/wiki/Content-addressable_storage)
+is a method of storing data in such a way that the address of the data is derived
+from the data itself. This means that if you store the same blob twice, it will
+be stored only once, and the address of the blob will be the same. This is achieved
+by using the [digest](https://en.wikipedia.org/wiki/Cryptographic_hash_function) of the blob as the address.
+
+This is the actual mechanism for allowing images to be optimal in terms of storage.
+Since most images begin by depending on the same base image, this means that
+it can be stored only once and referenced by multiple images.
+
+### Manifests
+
+A manifest is just a JSON document. It is stored seperately from the blobs.
+But similarly to the blobs, it is also content addressable. The manifest gets
+formated to a canonical form and then hashed. The hash becomes the address of
+the manifest.
+
+Where manifests become powerful is when they are used to reference to other
+manifests or blobs. These references are called [OCI Content Descriptors](https://github.com/opencontainers/image-spec/blob/v1.0.1/descriptor.md).
+For example, the Image Manifest that most readers would be familiar with containes
+a lisf of layers that are used to build the image. Each layer is a reference to
+a blob.
+
+An additional reference outside the layer references is used for storing 
+additional configuration. More info can be found in the official [OCI Image Manifest spec](https://github.com/opencontainers/image-spec/blob/v1.0.1/manifest.md).
+
+The term _artifact_ is just a generalization over the Image Manifest idea.
+The Image Manifest has a specific format that can be found in the aforementioned [OCI Image Manifest spec](https://github.com/opencontainers/image-spec/blob/v1.0.1/manifest.md).
+The important field is the .config.mediaType field. Based on the official guidelines,
+it can be used for defining types other than an image. More info can be found [here](https://github.com/opencontainers/image-spec/blob/main/manifest.md#guidelines-for-artifact-usage).
+
+Combining all of these ideas, we can see that artifacts can be represented by 1 
+specific object - a manifest. That manifest, based on the media type in the config, can be 
+interpreted by different programs for their specific use case. Eg. a Helm chart.
+Helm charts specifically have a mediaType of `application/vnd.cncf.helm.config.v1+json`.
+Layers, too, have a media type. The media type of a layer when the artifact
+is a helm chart is one of:
+- application/vnd.cncf.helm.config.v1+json
+- application/vnd.cncf.helm.chart.content.v1.tar+gzip
+- application/vnd.cncf.helm.chart.provenance.v1.prov
+
+More info specifically on Helm charts as OCI Artifacts can be found here: [Helm OCI MediaTypes](https://helm.sh/blog/helm-oci-mediatypes/).
+
+### Tags
+
+While manifests and blobs being content addressable is nice, it becomes hard
+to address them in a human-readable format. Tags allow us to add a reference
+to a given manifest that differs from the digest of the object.
+
+Normally, whenever we push an image to a registry, if we don't give it an
+explicit tag, the registry would most likey give it the tag `latest`. This,
+actually is not mandatory based on the distribution spec. Manifests don't 
+require that they have a tag.
+
+Multiple tags can point to the same manifest.
+
+### Cosign
+
+Cosign is generally a part of the bigger project called [sigstore](https://sigstore.dev/).
+Sigstore is an "open-source project for improving software supply chain security".
+It aims for developers to have a hassle-free way to sign their artifacts while
+leveraging other components that add additional layers of security to the
+signature process. Eg. Rekor and Fulcio.
+
+Cosign is a CLI tool aiming to tie all the components of sigstore together.
+In our specific case, we use it only as a format for creating signatures and 
+then verifying them. In theory, we wouldn't need to use Cosign for this, but
+it uses a popular format for the signatures and the libraries help us 
+save work.
+
--- a/pkg/lakom/resolvetag/admission_test.go
+++ b/pkg/lakom/resolvetag/admission_test.go
@@ -38,12 +38,6 @@ var (
 	apiReader *mockclient.MockReader
 )
 
-var _ = BeforeSuite(func() {
-	scheme = runtime.NewScheme()
-	err := corev1.AddToScheme(scheme)
-	Expect(err).ToNot(HaveOccurred())
-})
-
 var _ = Describe("Admission Handler", func() {
 	var (
 		ctx     = context.Background()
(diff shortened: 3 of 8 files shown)


rrhubenov:
@vpnachev Hi Vladi, I'd love if you have some ideas for how this setup can be improved. There is some code duplication that can be extracted and places where some 'hacks' were needed due to the way ginkgo works.

pkg/lakom/resolvetag/resolvetag_suite_test.go
```diff
@@ -5,13 +5,207 @@
...
+		Bytes: derBytes,
+	}
+
+	// Encode the PEM block to a string
+	pemBytes := pem.EncodeToMemory(pemBlock)
+
+	return string(pemBytes), nil
+}
+
+func writeRandomImage(tag string) (registryv1.Image, name.Reference, error) {
```
vpnachev, line 141: I would suggest to name this function `createImage` or `createTestImage`

pkg/lakom/resolvetag/resolvetag_suite_test.go
```diff
@@ -5,13 +5,207 @@
...
+	Expect(err).ToNot(HaveOccurred())
+	publicKey, err = publicKeyToPEM(privateKey.Public())
+	Expect(err).ToNot(HaveOccurred())
+
+	err = signImage(signedImage, signedImageFullRef)
+	Expect(err).ToNot(HaveOccurred())
+})
+
+var _ = AfterSuite(func() {
+	server.Close()
```
vpnachev, line 108: From the `DeferCleanup` docs

> When DeferCleanup is called in BeforeSuite, SynchronizedBeforeSuite, AfterSuite, or SynchronizedAfterSuite the registered callback will be invoked when the suite completes (i.e. it will behave like an AfterSuite node)

I think you can skip the `AfterSuite` node. Then you do not need the `server` variable to be in the global scope, it can be a local to the `BeforeSuite` node scope.

pkg/lakom/resolvetag/resolvetag_suite_test.go
```diff
@@ -5,13 +5,207 @@
...
+})
+
+func startFakeRegistry() (*url.URL, *httptest.Server) {
+	nopLog := log.New(io.Discard, "", 0)
```
vpnachev, line 112: It would be better to use the `GinkgoWriter` instead of `io.Discard`. In case of test failures, it is always better to have some logs. What I like about GinkgoWriter is that it is taking care to not write anything when the test succeeds, so it is not polluting the stdout when everything is working as expected. 

For consistency, I would propose to use `logzap` instead of `log` module
```go
logzap.New(logzap.WriteTo(GinkgoWriter))
```

pkg/lakom/resolvetag/resolvetag_suite_test.go
```diff
@@ -5,13 +5,207 @@
...
+	"github.com/sigstore/cosign/v2/pkg/oci/static"
+	"github.com/sigstore/sigstore/pkg/signature"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+var (
+	// Fake registry properties
+	registryURL *url.URL
+	server      *httptest.Server
```
vpnachev, line 41: Even thought this is just test code, it is nice to not use global variables when possible. 
`server` can stay local to the `BeforeSuite` node, and `registryURL` will be also defined in `BeforeSuite` and passed as function parameter where needed.

pkg/lakom/resolvetag/resolvetag_suite_test.go
```diff
@@ -5,13 +5,207 @@
...
+
+	signedImageTagRef      string
+	unsignedImageTagRef    string
+	nonExistantImageTagRef string
+
+	// Key for signing images in fake registry
+	privateKey *ecdsa.PrivateKey
+
+	// Public key in PEM format for verifying signatures in the fake registry
+	publicKey string
```
vpnachev, line 54: `publicKey` seems unused, it is just initialized. You can remove it.

pkg/lakom/resolvetag/resolvetag_suite_test.go
```diff
@@ -5,13 +5,207 @@
...
+
+	err = signImage(signedImage, signedImageFullRef)
+	Expect(err).ToNot(HaveOccurred())
+})
+
+var _ = AfterSuite(func() {
+	server.Close()
+})
+
+func startFakeRegistry() (*url.URL, *httptest.Server) {
```
vpnachev, line 111: I think you can skip `Fake` from the name, when it is in a test. 
From the code of `registry.New` it looks like a real registry to me (it is just not configured with persistent storage), i.e. it is not a fake implementation. 

```suggestion
func startRegistry() (*url.URL, *httptest.Server) {
```

pkg/lakom/resolvetag/resolvetag_suite_test.go
```diff
@@ -5,13 +5,207 @@
...
+
+	signedImageFullRef   string
+	unsignedImageFullRef string
+
+	signedImageTagRef      string
+	unsignedImageTagRef    string
+	nonExistantImageTagRef string
+
+	// Key for signing images in fake registry
+	privateKey *ecdsa.PrivateKey
```
vpnachev, line 51: `privateKey` also can be moved out from the global scope, it can be defined in the `BeforeSuite` and passed as parameter to `signImage`.

pkg/lakom/resolvetag/resolvetag_suite_test.go
```diff
@@ -5,13 +5,207 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = ociRemote.WriteSignatures(digest.Context(), si)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
vpnachev, line 210:
```suggestion
	return ociRemote.WriteSignatures(digest.Context(), si)
```

vpnachev (commented):
I still have to go through the changes in verifysignature module

pkg/lakom/verifysignature/verifier_test.go
```diff
@@ -64,8 +58,8 @@ IqozONbbdbqz11hlRJy9c7SG+hdcFl9jE9uE/dwtuwU2MqU9T/cN0YkWww==
 			PublicKeys: []config.Key{
 				{
 					Name:      "test",
-					Key:       cosignPublicKey,
-					Algorithm: config.RSAPKCS1v15SHA256,
+					Key:       publicKey,
+					Algorithm: "",
```
vpnachev, line 62: Not sure why is the algorithm unset, but as `Algorithm` is of type `string` it will be by default set to `""`, i.e. you do not have to explicitly set it to the empty string and can omit this line.

```suggestion
```

pkg/lakom/verifysignature/admission_test.go
```diff
@@ -29,58 +28,29 @@ import (
...
 	var (
-		ctx     = context.Background()
-		logger  logr.Logger
-		handler admission.Handler
-		pod     = &corev1.Pod{
+		ctx          = context.Background()
+		logger       logr.Logger
+		handler      admission.Handler
+		cosignConfig lakomconfig.Config
+		pod          = &corev1.Pod{
```
vpnachev, line 45:
```suggestion
		pod          *corev1.Pod
```

As the variable is initialized in `BeforeEach`, you should just declare it here.

pkg/lakom/verifysignature/verifier_test.go
```diff
@@ -81,16 +75,28 @@ IqozONbbdbqz11hlRJy9c7SG+hdcFl9jE9uE/dwtuwU2MqU9T/cN0YkWww==
...
+			// Thus when testing the fake registry on localhost, the `false` here does not matter.
 			directVerifier = verifysignature.NewDirectVerifier(*lakomConfig, false)
 		})
 
 		DescribeTable("Verify images",
-			func(image string, expectedVerificationResult, expectErr bool, errorMessage string) {
-				verified, err := directVerifier.Verify(ctx, image, kcr)
+			// The image variable is a pointer to a string because for some unknown reason
+			// when a string variable is passed to the `Entry` function, a copy of the string object gets used
+			// rather than the original object.
```
vpnachev, line 90: I remember I was facing this issue. Here the issue is in the `DescribeTable`, in practice ginkgo translates each `Entry` into an `It` container during the build, and it is copying the variables as they are. This happens before the `Before*` nodes, i.e. when the variables are yet not assigned with the desired values, hence the tests run with the variables assigned to the default value of their type. 

This is not an issue when `Entry` uses references, e.g. pointers, slices, interfaces, etc. as these even though the tests are built with nil references, before the tests are started the `Before*` nodes changes the values of these reference variables (not sure If I explained it properly, but I hope you get what I mean). 

Basically, it evaluate the variables during compile time and passes this values to the functions built out of `Entry`. Using a pointer allows you to change this value later during run time. 

I think this issue https://github.com/onsi/ginkgo/issues/378 describes exactly the problem and provides some ideas how to circumvent it.

pkg/lakom/verifysignature/verifier_test.go
```diff
@@ -121,12 +127,9 @@ IqozONbbdbqz11hlRJy9c7SG+hdcFl9jE9uE/dwtuwU2MqU9T/cN0YkWww==
 		var (
 			cache          verifysignature.SignatureVerificationResultCache
 			cachedVerifier verifysignature.Verifier
+			err            error
 		)
 		BeforeEach(func() {
```
vpnachev, line 132: The `err` variable is not used later in this scope, hence it is better to declare it only in the function passed to `BeforeEach`.


```suggestion
		)
		BeforeEach(func() {
		var err error
```

pkg/lakom/verifysignature/verifysignature_suite_test.go
```diff
@@ -5,13 +5,187 @@
 package verifysignature_test
 
 import (
+	"crypto"
```
vpnachev, line 8: In this file the feedback would be the same as the `resolvetag_suite_test.go`, thus I will skip it to avoid duplication.

docs/oci_registries_crash_course.md
```diff
@@ -0,0 +1,115 @@
...
+be stored only once, and the address of the blob will be the same. This is achieved
+by using the [digest](https://en.wikipedia.org/wiki/Cryptographic_hash_function) of the blob as the address.
+
+This is the actual mechanism for allowing images to be optimal in terms of storage.
+Since most images begin by depending on the same base image, this means that
+it can be stored only once and referenced by multiple images.
+
+### Manifests
+
+A manifest is just a JSON document. It is stored seperately from the blobs.
```
vpnachev, line 58:
```suggestion
A manifest is just a JSON document. It is stored separately from the blobs.
```

docs/oci_registries_crash_course.md
```diff
@@ -0,0 +1,115 @@
...
+But similarly to the blobs, it is also content addressable. The manifest gets
+formated to a canonical form and then hashed. The hash becomes the address of
```
vpnachev, line 60:
```suggestion
formatted to a canonical form and then hashed. The hash becomes the address of
```

docs/oci_registries_crash_course.md
```diff
@@ -0,0 +1,115 @@
...
+the manifest.
+
+Where manifests become powerful is when they are used to reference to other
+manifests or blobs. These references are called [OCI Content Descriptors](https://github.com/opencontainers/image-spec/blob/v1.0.1/descriptor.md).
+For example, the Image Manifest that most readers would be familiar with containes
```
vpnachev, line 65:
```suggestion
For example, the Image Manifest that most readers would be familiar with contains
```

docs/oci_registries_crash_course.md
```diff
@@ -0,0 +1,115 @@
...
+a lisf of layers that are used to build the image. Each layer is a reference to
```
vpnachev, line 66:
```suggestion
a list of layers that are used to build the image. Each layer is a reference to
```

docs/oci_registries_crash_course.md
```diff
@@ -0,0 +1,115 @@
...
+a blob.
+
+An additional reference outside the layer references is used for storing 
+additional configuration. More info can be found in the official [OCI Image Manifest spec](https://github.com/opencontainers/image-spec/blob/v1.0.1/manifest.md).
+
+The term _artifact_ is just a generalization over the Image Manifest idea.
+The Image Manifest has a specific format that can be found in the aforementioned [OCI Image Manifest spec](https://github.com/opencontainers/image-spec/blob/v1.0.1/manifest.md).
+The important field is the .config.mediaType field. Based on the official guidelines,
```
vpnachev, line 74:
```suggestion
The important field is the `.config.mediaType` field. Based on the official guidelines,
```

docs/oci_registries_crash_course.md
```diff
@@ -0,0 +1,115 @@
...
+More info specifically on Helm charts as OCI Artifacts can be found here: [Helm OCI MediaTypes](https://helm.sh/blog/helm-oci-mediatypes/).
+
+### Tags
+
+While manifests and blobs being content addressable is nice, it becomes hard
+to address them in a human-readable format. Tags allow us to add a reference
+to a given manifest that differs from the digest of the object.
+
+Normally, whenever we push an image to a registry, if we don't give it an
+explicit tag, the registry would most likey give it the tag `latest`. This,
```
vpnachev, line 96: Is automatically tagging with `latest` in the spec? I am not aware of registry that is doing this, usually it is a client side configuration (mainly on the CI instruments building and pushing the images to the registries).
It should not be required each image to have a tag, this is why I find it unexpected the registry to tag untagged image.


```suggestion
Normally, whenever we push an image to a registry, if we don't give it an
explicit tag, the registry would most likely give it the tag `latest`. This,
```

docs/oci_registries_crash_course.md
```diff
@@ -0,0 +1,115 @@
...
+actually is not mandatory based on the distribution spec. Manifests don't 
+require that they have a tag.
+
+Multiple tags can point to the same manifest.
+
+### Cosign
```
vpnachev, line 102: Do you plan to extend a bit the doc about cosign? 
It would be helpful to have example artifact with signature and to describe how the clients discover the where the signature is stored. As lakom is not supporting keyless signing, I would not add any more information about Fulcio and Rekor and their use case.

vpnachev (commented):
Thanks for the changes!
This will resolve our long standing technical debt with unit tests that are actually not unit tests and depend on external resources/services as well as network connectivity. 

I like the PR, we just need to polish the changes here and there.

pkg/lakom/resolvetag/admission_test.go
```diff
@@ -92,6 +86,20 @@ var _ = Describe("Admission Handler", func() {
 		Expect(err).ToNot(HaveOccurred())
 
 		handler = h
+
+		pod = &corev1.Pod{
```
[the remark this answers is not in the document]
rrhubenov: Sadly, that will cause a nil pointer dereference during the spec build of the tests that ginkgo goes through.
In the `Entries` in the `DescribeTable`  of the same file, the pod object is used to create the request object.
Since the calls to the `Entry` functions happen during the ginkgo spec tree build step, the pod object gets used before a `BeforeEach` can be called. 

We can avoid this by instead of passing a `admission.Request` object in the function for the `DescribeTable`, we pass the GVK, the operation type and the ptr to the object itself (the pod in this case). 

Ref of the code I'm talking about: https://github.com/gardener/gardener-extension-shoot-lakom-service/blob/6025d6e3442168011a4371e3ff4759bf9fd51ba2/pkg/lakom/resolvetag/admission_test.go#L97-L115

docs/oci_registries_crash_course.md
rrhubenov: I'd actually really like to that. I'm glad you find it a good idea, I'll add docs about the cosign signing & verification process.

pkg/lakom/resolvetag/resolvetag_suite_test.go
rrhubenov: Agree, I'd too prefer to avoid globals.

pkg/lakom/resolvetag/resolvetag_suite_test.go
rrhubenov: I was unable to use logzap since the type of the Logger is go-logr/logr.Logger, while the registry.New function
accepts the log/log.Logger type from the standard lib. I could not find a way to extract a log.Logger from the logr.Logger type.

Maybe you know a way?

pkg/lakom/verifysignature/verifier_test.go
rrhubenov: Thank you! I hadn't stumbled on this issue.
Although I'm not very satisfied with the workarounds :(. The workarounds are basically:
- Use some sort of reference to the object (ptr, index to the element in an array, etc.)
- Use a closure

Both have their pros and cons. Not sure whether one of them is better than the others.

pkg/lakom/resolvetag/admission_test.go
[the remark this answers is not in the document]
rrhubenov: I've simplified the pod reinitialisation to just update the `image` field with the new value in the `signedImageTagRef`. 

Maybe there's a cleaner way

pkg/lakom/verifysignature/admission_test.go
rrhubenov: I've applied the same fix as the previous occurrence of this issue

pkg/lakom/verifysignature/verifier_test.go
rrhubenov: I've left the workaround with pointers which I find relatively the same as the other workarounds. WDYT?

rrhubenov:
@vpnachev I really appreciate the thorough review. I've left comments on the parts that I think require more discussion

pkg/lakom/verifysignature/verifier_test.go
rrhubenov: As far as I can tell, the algorithm type here does not reflect heavily on the tests that we're trying to cover.
When the algorithm is not `RSA` the default hash of SHA256 is returned.

Would we want to include different keys and algorithms in our tests? I'd say it might not be useful for us

pkg/lakom/resolvetag/resolvetag_suite_test.go
vpnachev: My bad, I have not payed attention that the logr.Logger interface cannot be used. 
Then it is enough just to use the GinkgoWriter instead of io.Discard.

.

pkg/lakom/verifysignature/verifier_test.go
vpnachev: Most probably testing different algorithms should be done in e2e tests.

docs/oci_registries.md
```diff
@@ -113,3 +113,46 @@ then verifying them. In theory, we wouldn't need to use Cosign for this, but
...
+}
+```
+
+We can see that the hash of the manifest is stored in the `Docker-manifest-digest`.
+Now this object is the one that gets signed using our private key. The signed
+version of this objest is the one that gets stored and gets used for
+verification. 
+
+Storage of the signature is done via an image manifest that gets uploaded
+in the same repo as the image, but tagged with `<image-digest>.sig`.
```
vpnachev, line 148:
```suggestion
in the same repo as the image, but tagged with `sha256-<container-image-digest>.sig`.
```

pkg/lakom/resolvetag/resolvetag_suite_test.go
```diff
@@ -5,13 +5,173 @@
...
+
+	signedImageFullRef = signedImageRef.Name()
+	unsignedImageFullRef = unsignedImageRef.Name()
+
+	signedImageTagRef = signedImageRef.Context().Tag(signedImageTag).String()
+	unsignedImageTagRef = unsignedImageRef.Context().Tag(unsignedImageTag).String()
+
+	nonExistentImageTagRef = fmt.Sprintf("%s:nonexistant", signedImageRef.Context().Name())
+
+	privateKey, err := cosign.GeneratePrivateKey()
```
vpnachev, line 95: The function docs says
> // TODO(jason): Move this to an internal package.

Once it is moved, we will not be able to use it. I think it would better to directly generate the key using the `crypto/ecdsa.GenerateKey` (maybe it is possible to use `crypt/rsa.GenerateKey`, not sure which one was quicker and lighter).

pkg/lakom/resolvetag/resolvetag_suite_test.go
```diff
@@ -5,13 +5,173 @@
...
+	Expect(err).ToNot(HaveOccurred())
+
+	err = signImage(signedImage, signedImageFullRef, privateKey)
+	Expect(err).ToNot(HaveOccurred())
```
vpnachev, line 99:
```suggestion
	Expect(signImage(signedImage, signedImageFullRef, privateKey)).To(Succeed())
```

The other pattern is used when the function returns more than one result.

pkg/lakom/resolvetag/resolvetag_suite_test.go
```diff
@@ -5,13 +5,173 @@
...
+})
+
+func startRegistry() (*url.URL, *httptest.Server) {
+	ginkgoLogger := log.New(GinkgoWriter, "", 0)
+	s := httptest.NewServer(registry.New(registry.Logger(ginkgoLogger)))
+	u, err := url.Parse(s.URL)
+	if err != nil {
+		log.Fatal("Error parsing")
+	}
```
vpnachev, line 108:
```suggestion
	Expect(err).ToNot(HaveOccurred())
```

Exiting with log.Fatal seems too disruptive here and needs additional work to make the reason for the exit apparent. It is enough just to expect that no error has occurred.

pkg/lakom/verifysignature/verifier_test.go
vpnachev: You can remove the comments from here or just shorten them by putting reference to onsi/ginkgo#378

pkg/lakom/resolvetag/resolvetag_suite_test.go
rrhubenov: Yes, I remember seeing that comment and thinking this should be a temporary solution.
Good catch

vpnachev (approved):
/lgtm
closes #37

PR #143, opened 2025-02-18
Add support for projected kubeconfig in lakom chart

### LINKED ISSUE #142: Add support for projected kubeconfig in the lakom chart
Add support for projected kubeconfig in the lakom chart similar to the one already available for the admission controller https://github.com/gardener/gardener-extension-shoot-lakom-service/blob/c3fe2c3a7403d378e9ca3433a0b7be562398f33f/charts/gardener-extension-shoot-lakom-admission/charts/runtime/values.yaml#L27-L30

**Why is this needed**:
After #97 lakom can verify the `ControllerDeployment`s and `Gardenlet`s in the virtual garden (target) cluster , but it has to run in the runtime (source) cluster, hence it needs to mount a generic kubeconfig and token secrets that will grant it with access to the target cluster by leveraging [token requestor controller](https://github.com/gardener/gardener/blob/master/docs/concepts/resource-manager.md#tokenrequestor-controller)

/area ipcei
/area security

vpnachev:
Add support for projected kubeconfig in lakom chart.
The PR also contains a bugfix in the validating webhook configuration URL path and small improvements in the gardener-extension-shoot-lakom-admission chart to not use both kubeconfig and projectedKubeconfig

Fixes #142 

**Release note**:

```feaeture operator
`lakom` helm chart now supports projected kubeconfig, this allows lakom to run in a source cluster and work for a target cluster, while the secrets with the kubeconfig and the token are externally managed.
```

```bugfix operator
Fix a bug when the `lakom` helm chart is installed with the value `.Values.admissionConfig.clientConfig.urlHostname` set the URL path in the ValidatingWebhookConfiguration was wrongly set to `/lakom/resolve-tag-to-digest` instead of `/lakom/verify-cosign-signature`.
```

### DIFF
--- a/charts/gardener-extension-shoot-lakom-admission/charts/runtime/templates/deployment.yaml
+++ b/charts/gardener-extension-shoot-lakom-admission/charts/runtime/templates/deployment.yaml
@@ -35,7 +35,7 @@ spec:
       priorityClassName: {{ .Values.gardener.runtimeCluster.priorityClassName }}
       {{- end }}
       serviceAccountName: {{ include "name" . }}
-      {{- if .Values.kubeconfig }}
+      {{- if or .Values.kubeconfig .Values.projectedKubeconfig }}
       automountServiceAccountToken: false
       {{- end }}
       containers:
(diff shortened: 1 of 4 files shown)


rrhubenov:
Is the flag going to be implemented in an additional PR?

vpnachev:
Which flag?

rrhubenov:
Sorry, the `--kubeconfig` flag.

vpnachev:
Ah, I abandoned the idea to implement support for multiple clusters in single lakom instance just to cover the garden runtime and virtual cluster, instead we will go with additional instance of lakom that runs in the runtime cluster but works for virtual one only. Therefore it will not need new `kubeconfig` flag.

rrhubenov (approved):
/lgtm

PR #144, opened 2025-02-18
Split lakom chart into application and runtime subcharts

vpnachev:
Split lakom chart into application and runtime subcharts. This is needed for easier installation of the lakom admission controller in the Garden runtime cluster for the Garden virtual cluster, i.e. the runtime subchart is installed in the runtime cluster and the application subchart is installed in the virtual cluster. When lakom is installed in the cluster it is targeting, both application and runtime subcharts have to be installed into this cluster.

Fixes #

**Release note**:

```other operator
The `lakom` helm chart is now split into `runtime` and `application` subcharts to allow easier installation into source and target clusters, when it is running one cluster but admitting requests for another cluster.
```

```breaking operaotr
The `lakom` chart values structure has been changed in the following way:
- `.Values.name` property is moved to `.Values.global.name`
- `.Values.kubeconfig` property is moved to `.Values.global.kubeconfig`
- `.Values.admissionConfig` property is moved to `.Values.application.admissionConfig`
- all other values properties are moved under `.Values.runtime`.
Please, adapt accordingly your values files.
```


charts/lakom/charts/runtime/templates/deployment.yaml
```diff
@@ -52,21 +52,21 @@ spec:
...
-      {{- if or .Values.kubeconfig .Values.projectedKubeconfig }}
+      {{- if or .Values.global.kubeconfig .Values.projectedKubeconfig }}
       automountServiceAccountToken: false
       {{- else }}
-      serviceAccountName: {{ .Values.name }}
+      serviceAccountName: {{ .Values.global.name }}
       {{- end }}
       containers:
-      - name: {{ .Values.name }}
+      - name: lakom-admission-controller
```
rrhubenov, line 69:
```suggestion
      - name: {{ .Values.global.name }}
```
Is there a reason the name is not used?

charts/lakom/values.yaml
rrhubenov, line 1: Since the values for the runtime and application charts are already present in their corresponding `values.yalm` files, is there a reason for them to be repeated in this parent `values.yaml`?

rrhubenov (commented):
Looks good, just some questions

charts/lakom/charts/runtime/templates/deployment.yaml
vpnachev: Yes, there is a reason. The container name is usually part of the prometheus metrics and it would be useful to easily filter the metrics from all lakom containers in a cluster, this is why I changed the container name to a static name.

charts/lakom/values.yaml
vpnachev: Looking at https://helm.sh/docs/topics/charts/#the-chart-file-structure, values.yaml is not described as optional file and it must be provided. When we have it, it makes sense to have default values set even though we are not intending to install from the subcharts but always from the root chart, I do not thing we can make this requirement mandatory and some users still may install the solely a subchart. 

TL;DR - it is nice to have some default values so that users have a single place where they can see what can be configured. Having a superchart combining two subcharts requires it's own values.yaml which is basically the result of merging the subcharts' values.yaml files that lead to a duplication.

charts/lakom/charts/runtime/templates/deployment.yaml
rrhubenov: So by leaving it static you're assuring that it won't change In the future by mistake?

charts/lakom/values.yaml
rrhubenov: So in other words, the values.yaml file in the subcharts could be made empty, but then we'd restrict installation of just the subcharts. Am I correct?

charts/lakom/charts/runtime/templates/deployment.yaml
vpnachev: Yes

rrhubenov: Thank you for the explanation

charts/lakom/values.yaml
vpnachev: Actually, today I just faced an issue that the default values in the subchart's values.yaml cannot be unset. (I wanted to unset the namespaceSelector). It turns out a fix (https://github.com/helm/helm/pull/12879) was proposed a year ago and it was recently merged and released with helm v3.17.1, but I found multiple other issues related to management of the subchart values, therefore decided to remove everything but `global` section from these subchart values.

rrhubenov: That's sad to hear. Hopefully there'll be improvements on this topic in the next versions

rrhubenov (approved):
/lgtm

PR #146, opened 2025-03-05
Fix lakom to consider algorithm when making trusted keys unique

vpnachev:
Fix lakom to consider algorithm when making trusted keys unique

Fixes #

/kind bug

**Release note**:

```bugfix operator
Lakom admission controller is now considering two trusted keys different not only when the actual crypto keys are different but also when they are configured with different algorithms.
```


dimityrmirchev (approved):
/lgtm

PR #152, opened 2025-03-24
Add support for image pull secrets used in ControllerDeployments, Extensions and Gardenlets

vpnachev:
**What this PR does / why we need it**:
Add support for image pull secrets used in ControllerDeployments, Extensions and Gardenlets

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:
x-ref: https://github.com/gardener/gardener/pull/11324

**Release note**:

```feature operator
Lakom admission controller now can verify `ControllerDeployment`s, `Gardenlet`s and `Extension`s that are using private OCI registries for their helm charts.
```

```other operator
The following 3rd party dependencies have been updated:
- github.com/gardener/gardener v1.110.1 -> v1.115.0
- golang.org/x/tools v0.28.0 -> v0.31.0
- k8s.io/api v0.31.3 -> v0.32.3
- k8s.io/apimachinery v0.31.3 -> v0.32.3
- k8s.io/autoscaler/vertical-pod-autoscaler v1.2.1 -> v1.2.2
- k8s.io/client-go v0.31.3 -> v0.32.3
- k8s.io/code-generator v0.31.3 -> v0.32.3
- k8s.io/component-base v0.31.3 -> v0.32.3
- sigs.k8s.io/controller-runtime v0.19.3 -> v0.20.3
```

PR #153, opened 2025-03-24
Enable and configure `importas` linter

vpnachev:

Fixes #

To be merged after #152 

**Release note**:

```other operator

```


.golangci.yaml
```diff
@@ -22,3 +23,220 @@ issues:
   - "exported: (type|func) name will be used as .* by other packages, and that stutters;"
   exclude-files:
   - "zz_generated\\..*\\.go$"
+
+linters-settings:
```
rrhubenov, line 27: Are these values taken from somewhere?

rrhubenov (approved):
/lgtm
Cool addition, just one question

.golangci.yaml
vpnachev: Yes, they are taken from https://github.com/gardener/gardener/blob/7361a19f4c3830a9f5134c073d3bfd72f4dcfa49/.golangci.yaml.in#L66-L271. I will have to think about whether we can import them with the option to provide custom ones or overwrite some of the entries.

vpnachev: [8f37e4d](https://github.com/gardener/gardener-extension-shoot-lakom-service/pull/153/commits/8f37e4dadae6902b35b0f00ada4851bc74ec88b1) merges the importas alias config with the one from Gardener. 21bf62c65674 allows to overwrite the gardener config.

rrhubenov: Nice, this is definitely better :)

rrhubenov (approved):
/lgtm

PR #162, opened 2025-04-24
Fix extension controller permissions to deletecollection secrets

vpnachev:
Fix extension controller permissions to deletecollection secrets

Follow up on #155

**Release note**:

```bugfix operator
Fix a bug where extension controller was missing permission to `deletecollection` secrets.
```


rrhubenov (approved):
/lgtm

If I understand correctly, with #155 the verbs were explicitly listed rather than using a wildcard. Not sure where the controller is issuing a `deletecollection` command, though

vpnachev:
`deletecollection` permissions are required because of https://github.com/gardener/gardener-extension-shoot-lakom-service/blob/33ab30fbd857da56ca55a073fc280dcdfbd2a131/pkg/controller/lifecycle/actuator.go#L247

PR #165, opened 2025-05-19
Restrict Workflow permissions to read-only

vpnachev:
Restrict Workflow permissions to read-only

Fixes #

**Release note**:

```other operator

```


rrhubenov (approved):
/lgtm

Seems like the two steps only need permissions to read the contents of the repo, so this should be fine

PR #192, opened 2025-09-01
Fix the admission runtime name

dimitar-kostadinov:
Fix the name of the admission runtime chart.

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -88,7 +88,7 @@ jobs:
                 attribute: image.repository
               - ref: ocm-resource:gardener-extension-shoot-lakom-admission.tag
                 attribute: image.tag
-          - name: shoot-lakom-admission-application-runtime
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -88,7 +88,7 @@ jobs:
                 attribute: image.repository
               - ref: ocm-resource:gardener-extension-shoot-lakom-admission.tag
                 attribute: image.tag
-          - name: shoot-lakom-admission-application-runtime
+          - name: shoot-lakom-admission-runtime
```
vpnachev, line 91: Thanks for opening the fix!

I have already proposed such change with https://github.com/gardener/gardener-extension-shoot-lakom-service/pull/189/commits/5ccae48aa6b39b0cca859ae5f7abcca5c489a404. 

I would propose to close this PR in favor of #189, what do you think?

PR #207, opened 2025-10-02
Fix Gardenlet tag resolver to not inject repository into the tag

vpnachev:
Fix Gardenlet tag resolver to not inject repository into the tag

Fixes #

**Release note**:

```bugfix operator
A bug in the `Gardenlet` image tag resolver, where the resolved image is used as tag instead of just the digest, has been fixed. 
```

/kind bug


dimityrmirchev (approved):
/lgtm

PR #227, opened 2025-11-11
Review RBAC permissions and security context of components

vpnachev:
Review RBAC permissions and security context of components

/area security
/kind enhancement 

Fixes #

**Release note**:

```other operator
NONE
```


charts/gardener-extension-shoot-lakom-service/templates/deployment.yaml
```diff
@@ -64,7 +64,6 @@ spec:
       serviceAccountName: gardener-extension-{{ .Values.serviceName }}
       securityContext:
         runAsNonRoot: true
-        allowPrivilegeEscalation: false
```
dimityrmirchev: This field seems like it is not part of the pod security context. How was it working before?

PR #264, opened 2026-02-12
Switch `.spec.trafficDistribution` field to `PreferSameZone` for Kubernetes 1.34+

Kostov6:

/area networking
/area high-availability
/kind enhancement

This PR adapts the topogy-ware `Service's` `.spec.trafficDistribution` field from `PreferClose` to `PreferSameZone` for Kubernetes 1.34+.

Part of https://github.com/gardener/gardener/issues/13942

Tested like this:
```
❯ helm template foo ./charts/gardener-extension-shoot-lakom-admission/charts/runtime -s templates/service.yaml --set service.topologyAwareRouting.enabled=true --kube-version 1.33.0 | grep trafficDistribution
  trafficDistribution: PreferClose

❯ helm template foo ./charts/gardener-extension-shoot-lakom-admission/charts/runtime -s templates/service.yaml --set service.topologyAwareRouting.enabled=true --kube-version 1.34.0 | grep trafficDistribution
  trafficDistribution: PreferSameZone
```

**Release note**:

```other operator
The `.spec.trafficDistribution` field of the topology-aware Services of the `shoot-lakom-admission` component will be automatically switched from the deprecated `PreferClose` to the new `PreferSameZone` option for Kubernetes 1.34+.
```


ialidzhikov:
> What about the other Service resources maintained by the component, shouldn't they also make use of the `trafficDistribution` feature?

@vpnachev , thanks for the pointer. But you would also agree that this PR is about adapting existing topology-aware Services to the new way of making a service topology-aware in Kubernetes 1.34. It is not about going over all Services and re-evaluating if they need to be topology-aware or not. This is another workstream.
I would say that this is rather up to the extension owners to decide. There are guidelines provided in [Component Checklist](https://github.com/gardener/gardener/blob/v1.136.0/docs/development/component-checklist.md) and in [Topology-Aware Traffic Routing](https://github.com/gardener/gardener/blob/v1.136.0/docs/operations/topology_aware_routing.md) when to make a Service topology-aware.

ialidzhikov:
/retitle Switch `.spec.trafficDistribution` field to `PreferSameZone` for Kubernetes 1.34+

vpnachev:
I have opened #268 to enable topology aware routing for the other services where I think it makes sense.

PR #268, opened 2026-02-16
Enable topology aware routing for more services

vpnachev:
/area networking
/kind enhancement

Enable topology aware routing for more services
- lakom service for shoots in the control plane
- lakom service for seeds in the kube-system namespace
- lakom service for lakom-admission-controller installed via the helm chart

https://github.com/gardener/gardener-extension-shoot-lakom-service/pull/264#pullrequestreview-3790081353

**Release note**:

```other operator
Topology aware routing has been conditionally enabled for the services exposing the lakom admission controller covering the shoot scenario, seed scenario, and when the controller is installed via the `lakom` helm chart.
```

### DIFF
--- a/pkg/controller/lifecycle/actuator.go
+++ b/pkg/controller/lifecycle/actuator.go
@@ -178,7 +178,7 @@ func (a *actuator) Reconcile(ctx context.Context, logger logr.Logger, ex *extens
 		a.serviceConfig.UseOnlyImagePullSecrets,
 		a.serviceConfig.AllowUntrustedImages,
 		a.serviceConfig.AllowInsecureRegistries,
-		isTopologyAwareRoutingForShootControlPlaneEnabled(cluster),
+		v1beta1helper.IsTopologyAwareRoutingForShootControlPlaneEnabled(cluster.Seed, cluster.Shoot),
 		*cluster.Seed.Status.KubernetesVersion,
 	)
 	if err != nil {
@@ -800,14 +800,4 @@ func getRoleBinding(scope lakom.ScopeType, shootAccessServiceAccountName string)
 		RoleRef:  roleRef,
 		Subjects: subjects,
 	}
-}
-
-func isTopologyAwareRoutingForShootControlPlaneEnabled(cluster *extensions.Cluster) bool {
-	var (
-		seed  = cluster.Seed
-		shoot = cluster.Shoot
-	)
-
-	return seed != nil && shoot != nil &&
-		v1beta1helper.IsTopologyAwareRoutingForShootControlPlaneEnabled(seed, shoot)
 }
--- a/pkg/controller/seed/add.go
+++ b/pkg/controller/seed/add.go
@@ -13,7 +13,7 @@ import (
 	"github.com/Masterminds/semver/v3"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/discovery"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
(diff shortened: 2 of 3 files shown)


pkg/controller/seed/add.go
```diff
@@ -33,13 +36,33 @@ type AddOptions struct {
...
+	// SeedTopologyAwareRoutingEnabled determines whether the seed topology aware routing is enabled or not.
+	SeedTopologyAwareRoutingEnabled bool
 }
 
 // AddToManager adds a Lakom Service seed bootstrap controller to the given Controller Manager.
 func AddToManager(_ context.Context, mgr manager.Manager) error {
+	clientset, err := kubernetes.NewForConfig(mgr.GetConfig())
+	if err != nil {
+		return fmt.Errorf("could not create Kubernetes clientset: %w", err)
+	}
```
ialidzhikov: You could use a `discovery.DiscoveryClient` instead of a full `kubernetes.ClientSet`. You are only using it to discover the Seed Kubernetes version.

You don't need to initialize a `kubernetes.ClientSet` which will discover all APIs and initialize a client for all built-in API resources.

Suggestion:
```diff
diff --git a/pkg/controller/seed/add.go b/pkg/controller/seed/add.go
index df75e58a..b658aef5 100644
--- a/pkg/controller/seed/add.go
+++ b/pkg/controller/seed/add.go
@@ -13,7 +13,7 @@ import (
 	"github.com/Masterminds/semver/v3"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/discovery"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
@@ -42,12 +42,12 @@ type AddOptions struct {
 
 // AddToManager adds a Lakom Service seed bootstrap controller to the given Controller Manager.
 func AddToManager(_ context.Context, mgr manager.Manager) error {
-	clientset, err := kubernetes.NewForConfig(mgr.GetConfig())
+	discoveryClient, err := discovery.NewDiscoveryClientForConfig(mgr.GetConfig())
 	if err != nil {
-		return fmt.Errorf("could not create Kubernetes clientset: %w", err)
+		return fmt.Errorf("could not create a discovery client: %w", err)
 	}
 
-	k8sVersionInfo, err := clientset.Discovery().ServerVersion()
+	k8sVersionInfo, err := discoveryClient.ServerVersion()
 	if err != nil {
 		return err
 	}

```

pkg/controller/seed/add.go
```diff
@@ -33,13 +36,33 @@ type AddOptions struct {
...
+
+	k8sVersionInfo, err := clientset.Discovery().ServerVersion()
+	if err != nil {
+		return err
+	}
+
+	k8sVersion, err := semver.NewVersion(k8sVersionInfo.GitVersion)
+	if err != nil {
+		return err
```
ialidzhikov: You could wrap the error.

pkg/controller/lifecycle/actuator.go
```diff
@@ -779,3 +801,13 @@ func getRoleBinding(scope lakom.ScopeType, shootAccessServiceAccountName string)
 		Subjects: subjects,
 	}
 }
+
+func isTopologyAwareRoutingForShootControlPlaneEnabled(cluster *extensions.Cluster) bool {
+	var (
+		seed  = cluster.Seed
+		shoot = cluster.Shoot
+	)
```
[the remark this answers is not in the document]
vpnachev: https://github.com/gardener/gardener-extension-shoot-lakom-service/blob/131fe2046f39852cb72a368c0d11db441f8a49f9/pkg/controller/lifecycle/actuator.go#L140-L142 

already checks if the `seed` is set.

https://github.com/gardener/gardener-extension-shoot-lakom-service/blob/131fe2046f39852cb72a368c0d11db441f8a49f9/pkg/controller/lifecycle/actuator.go#L125 rely that `shoot` is set, so it is indeed safe to remove the check for nil pointer.

PR #306, opened 2026-05-05
Drop support for K8s `<= 1.31`

timuthy:

/area open-source
/kind cleanup

This PR removes the support for Kubernetes versions <= 1.31.

Part of https://github.com/gardener/gardener/issues/13919

**Release note**:

```breaking operator
⚠️ This extension no longer supports Kubernetes versions `<= 1.31`. Please make sure to upgrade all Garden, Seed and Shoot clusters to at least version 1.32 before deploying this extension version.
```

### DIFF
--- a/charts/gardener-extension-shoot-lakom-admission/charts/runtime/templates/service.yaml
+++ b/charts/gardener-extension-shoot-lakom-admission/charts/runtime/templates/service.yaml
@@ -10,14 +10,8 @@ metadata:
   namespace: {{ .Release.Namespace }}
   annotations:
     networking.resources.gardener.cloud/from-all-webhook-targets-allowed-ports: '[{"protocol":"TCP","port":{{ .Values.webhookConfig.serverPort }}}]'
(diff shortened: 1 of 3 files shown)


theoddora (commented):
/lgtm

vpnachev (commented):
Thanks for the housekeeping activity!

Code-wise looks good to me, I am just wondering whether we should hold for some time this change, e.g. for as long as a supported Gardener version still supports k8s clusters of version 1.31?

timuthy:
> Code-wise looks good to me, I am just wondering whether we should hold for some time this change, e.g. for as long as a supported Gardener version still supports k8s clusters of version 1.31?

This might be a good point, in reality however, I'm not sure if it really makes a difference and is that important. There is only one Gardener version left `v1.141.x` that supports Kubernetes 1.31.
In addition, the `gardener/gardener` dependency was already bumped to `v1.142` and thus support for K8s 1.31 was already partly removed.

PR #304, opened 2026-04-30
Switch the default scope to `KubeSystem`

mimiteto:
**How to categorize this PR?**

/area compliance
/kind api-change

**What this PR does / why we need it**:
PR switches the default scope to 'KubeSystem', which, by default, forces all images in the `kube-system` to be validated, which increases confidence in the "system" workloads.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:
/invite @vpnachev 

**Release note**:

```breaking operator
The default scope for `lakom` is changed to 'KubeSystem'
```

PR #307, opened 2026-05-07
Make default lakom scope configurable 

vpnachev:
/area security
/kind enhancement

Make default lakom scope configurable via the extension controller configuration `defaultLakomScope`.

Replaces #304

cc @mimiteto @JordanJordanov 

**Release note**:

```feature operator
Extension configuration feature new field `defaultLakomScope` which can be used to overwrite the default lakom scope, the default behavior of using the scope `KubeSystemManagedByGardener` is preserved.
```


pkg/apis/config/v1alpha1/types.go
```diff
@@ -35,6 +37,9 @@ type Configuration struct {
 	AllowUntrustedImages bool `json:"allowUntrustedImages"`
 	// AllowInsecureRegistries allows Lakom to use HTTP for communication with the registries
 	AllowInsecureRegistries bool `json:"allowInsecureRegistries"`
+	// DefaultLakomScope is configuration exposing the option to overwrite the default scope
+	// for the lakom admission controller managed by the extension controller.
+	DefaultLakomScope lakom.ScopeType `json:"defaultLakomScope"`
```
dimityrmirchev: I only wonder if we should rename this to "defaultVerificationScope", "defaultScope" or something else. I am also fine with leaving it as is.

pkg/controller/lifecycle/actuator.go
```diff
@@ -110,14 +124,11 @@ func (a *actuator) Reconcile(ctx context.Context, logger logr.Logger, ex *extens
...
-			logger.Error(err, "Could not decode provider config. Using default value `KubeSystemManagedByGardener` for scope")
+			return fmt.Errorf("could not decode provider config, err: %w", err)
 		}
 	}
-	if lakomProviderConfig.Scope == nil {
-		logger.Info("No scope specified. Using default value `KubeSystemManagedByGardener` for scope")
-		lakomProviderConfig.Scope = ptr.To(lakom.KubeSystemManagedByGardener)
-	}
+
+	resolveScope(logger, lakomProviderConfig, a.serviceConfig.DefaultLakomScope)
```
dimityrmirchev: I wonder if we should do an assignment to make it clear that we are overwriting the config, something like

```golang
lakomProviderConfig.Scope, msg := resolveScope(...)
logger.LogInfo(msg, ...)
```

I am also fine with this implementation

dimityrmirchev (commented):
Two optional comment

pkg/apis/config/v1alpha1/types.go
vpnachev: Renamed to `defaultAdmissionScope` as it affects not only the validating webhook, but also the mutating one responsible to resolve image tags to digests.

pkg/controller/lifecycle/actuator.go
vpnachev: Changed the function to return an overwrite or the default scope, then this value is assigned to `lakomProviderConfig.Scope` only if it is unset.

dimityrmirchev (commented):
/lgtm
/approve

PR #332, opened 2026-07-14
Add support for Garden Extensions of class garden

### LINKED ISSUE #148: Support Garden Extensions, e.g. class=garden
[Garden resource](https://github.com/gardener/gardener/blob/master/docs/concepts/operator.md#garden-resources) has support of extensions, they will be deployed in the runtime cluster and be of class [`garden`](https://github.com/gardener/gardener/blob/196a6d6737738b62a8ec3bbc62cff4a03d88e1c8/pkg/apis/extensions/v1alpha1/types_defaults.go#L34-L35). `shoot-lakom-service` controller has to add support for this new class of extensions. 

`shoot-lakom-admission` controller can be extended to verify the Gardens.spec.extensions[@.type==shoot-lakom-service] similarly to how it is doing it for shoots already. 

**Why is this needed**:
To have a common convenient mechanism for installation of the `lakom` admission controller in the Garden cluster that will take care to verify:
- pod's images and operator.gardener.cloud/v1alpha1.Extensions in the runtime cluster
- gardenlets and controllerdeployments in the virtual cluster

h-mavrodiev:

/kind enhancement
/area usability

Introduces a common convenient mechanism for installation of the lakom admission controller in the Garden cluster that will take care to verify:

pod's images and operator.gardener.cloud/v1alpha1.Extensions in the runtime cluster
gardenlets and controllerdeployments in the virtual cluster

Fixes #148

/cc @vpnachev 

**Release note**:

```other operator
Lakom now supports extensions of class garden.
```


charts/gardener-extension-shoot-lakom-service/templates/rbac.yaml
```diff
@@ -18,6 +18,14 @@ rules:
...
   - list
   - watch
+- apiGroups:
+  - operator.gardener.cloud
+  resources:
+  - gardens
+  verbs:
+  - get
+  - list
+  - watch
```
vpnachev, line 30: These permissions should be granted only when the extension is installed for the garden class scenario.

charts/gardener-extension-shoot-lakom-service/templates/rbac.yaml
```diff
@@ -90,15 +98,61 @@ rules:
...
+  - coordination.k8s.io
+  resources:
+  - leases
+  resourceNames:
+  - {{ include "leaderelectionid" . }}
+  - gardener-extension-heartbeat
+  verbs:
+  - get
+  - patch
+  - update
```
vpnachev, line 129: Those are already granted via the ClusterRole, why is it necessary to grant them also via Role? 

Considering the leases are just several and resides in particular namespace, probably it is better to control the access to them via Role, but we should remove the duplications with the ClusterRole.

charts/gardener-extension-shoot-lakom-service/templates/rbac.yaml
```diff
@@ -90,15 +98,61 @@ rules:
...
+    app.kubernetes.io/name: gardener-extension-{{ .Values.serviceName }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: gardener-extension-{{ .Values.serviceName }}-leader-election
+subjects:
+- kind: ServiceAccount
+  name: gardener-extension-{{ .Values.serviceName }}
+  namespace: {{ .Release.Namespace }}
```
vpnachev: Please, keep the last new line in the files.

local-setup/operator/extension.yaml
```diff
@@ -20,6 +20,12 @@ spec:
...
       policy: Always
+      values: {}
+      runtimeClusterValues: {}
   resources:
   - kind: Extension
     type: shoot-lakom-service
+    clusterCompatibility:
+      - garden
+    autoEnable:
+    - garden
```
vpnachev: The local setup must cover both shoot and garden cases, also let's not enable it automatically for neither of the classes. 

```suggestion
    clusterCompatibility:
      - garden
      - shoot

```

pkg/constants/constants.go
```diff
@@ -15,12 +15,32 @@ const (
...
 	// ManagedResourceNamesShoot is the name used to describe the managed shoot resources.
 	ManagedResourceNamesShoot = ExtensionServiceName + "-shoot"
+	// ManagedResourceNamesGardenRuntime is the name used to describe the managed resources deployed on the runtime cluster for the garden extension class.
+	ManagedResourceNamesGardenRuntime = ExtensionServiceName + "-garden-runtime"
+	// ManagedResourceNamesGardenVirtual is the name used to describe the managed resources deployed on the virtual garden cluster for the garden extension class.
+	ManagedResourceNamesGardenVirtual = ExtensionServiceName + "-garden-virtual"
+	// ManagedResourceNamesGardenRuntimeWebhook is the name of the seed ManagedResource carrying the runtime garden webhook configs (registered in the runtime cluster).
+	ManagedResourceNamesGardenRuntimeWebhook = ExtensionServiceName + "-garden-runtime-webhook"
+	// ManagedResourceNamesGardenVirtualShoot is the name of the shoot ManagedResource carrying the virtual garden webhook configs (registered in the virtual garden).
+	ManagedResourceNamesGardenVirtualShoot = ExtensionServiceName + "-garden-virtual-shoot"
```
vpnachev: `*-shoot` is not appropriate name. Ideally, this should be named `virtual`, but considering we have two distinct deployment (one for the `runtime` and one for the `virtual` cluster) now using for the second time `virtual` term in the name would make it more confusing. Probably `-garden-virtual-config` is a better option.

vpnachev (commented):
Part one, I will continue tomorrow.

pkg/controller/lifecycle/actuator_test.go
```diff
@@ -55,6 +56,26 @@ var _ = Describe("Actuator", func() {
 		Expect(appPartOf).To(Equal("shoot-lakom-service"))
 	})
 
+	DescribeTable("Should get the garden pod labels with a variant-specific instance label",
+		func(virtualGarden bool, expectedInstance string) {
```
vpnachev:
```suggestion
		func(virtualGarden bool) {
```

The `expectedInstance` can be derived from `virtualGarden`, no need to pass it as parameter.

pkg/controller/lifecycle/actuator.go
```diff
@@ -262,16 +399,81 @@ func (a *actuator) delete(ctx context.Context, logger logr.Logger, ex *extension
...
+
+	if err := managedresources.WaitUntilDeleted(timeoutSeedCtx, a.client, namespace, constants.ManagedResourceNamesGardenRuntime); err != nil {
+		return err
+	}
+
+	if err := managedresources.DeleteForSeed(ctx, a.client, namespace, constants.ManagedResourceNamesGardenVirtual); err != nil {
+		return err
+	}
+
+	if err := managedresources.WaitUntilDeleted(timeoutSeedCtx, a.client, namespace, constants.ManagedResourceNamesGardenVirtual); err != nil {
```
vpnachev: Here `managedresources.WaitUntilDeleted` should be called each with own context with just initialized timeouts, otherwise all three deletions are assigned the same 2 minutes timeout.

pkg/controller/lifecycle/actuator.go
```diff
@@ -262,16 +399,81 @@ func (a *actuator) delete(ctx context.Context, logger logr.Logger, ex *extension
...
+		return err
+	}
+
+	if err := a.client.DeleteAllOf(ctx, &corev1.Secret{}, client.InNamespace(namespace), client.MatchingLabels(getLabels())); err != nil {
+		return err
+	}
+
 	if skipSecretManagerSecrets {
```
vpnachev: This is needed for the shoot's [control plane migration](https://github.com/gardener/gardener/blob/master/docs/extensions/migration.md), for Gardens there is no such operation and it is not needed.

pkg/controller/lifecycle/actuator.go
```diff
@@ -262,16 +399,81 @@ func (a *actuator) delete(ctx context.Context, logger logr.Logger, ex *extension
...
 		return nil
 	}
 
-	secretsManager, err := extensionssecretsmanager.SecretsManagerForCluster(ctx, logger.WithName("secretsmanager"), clock.RealClock{}, a.client, cluster, secrets.ManagerIdentity, nil)
+	garden, err := a.getGarden(ctx)
+	if err != nil {
+		return err
+	}
+
+	secretsManager, err := extensionssecretsmanager.SecretsManagerForGarden(ctx, logger.WithName("secretsmanager"), clock.RealClock{}, a.client, garden, secrets.ManagerIdentityGarden, nil, namespace)
```
vpnachev: When the secrets are deployed, the secrets manager is configured also with the `constants.LakomSystemNamespace` namespace, but it is not during the deletion. Why is this so?

pkg/controller/lifecycle/actuator.go
```diff
@@ -291,14 +493,25 @@ func (a *actuator) Restore(ctx context.Context, logger logr.Logger, ex *extensio
...
+		if err := managedresources.SetKeepObjects(ctx, a.client, ex.GetNamespace(), constants.ManagedResourceNamesShoot, true); err != nil {
+			return err
+		}
+		return a.deleteShoot(ctx, logger, ex, true)
+	case extensionsv1alpha1.ExtensionClassGarden:
+		// Keep objects for shoot virtual garden managed resources so that they are not deleted during the migration -- Is this even necessary?
+		if err := managedresources.SetKeepObjects(ctx, a.client, ex.GetNamespace(), constants.ManagedResourceNamesGardenVirtualWebhook, true); err != nil {
+			return err
+		}
+		return a.deleteGarden(ctx, logger, ex, true)
```
vpnachev:
```suggestion
```

pkg/controller/lifecycle/lakom_resources.go
```diff
@@ -0,0 +1,851 @@
...
+		"--port=" + serverPort.String(),
+	}
+	if opts.authMode == authInjectedKubeconfig {
+		args = append(args, "--kubeconfig="+gardenerutils.PathGenericKubeconfig)
+	}
+	args = append(args,
+		"--use-only-image-pull-secrets="+strconv.FormatBool(opts.useOnlyImagePullSecrets),
+		"--insecure-allow-untrusted-images="+strconv.FormatBool(opts.allowUntrustedImages),
+		"--insecure-allow-insecure-registries="+strconv.FormatBool(opts.allowInsecureRegistries),
+	)
```
vpnachev: You can firstly define all required flags and then append the optional one(s).

```suggestion
	args := []string{
		"--cache-ttl=" + cacheTTL.String(),
		"--cache-refresh-interval=" + cacheRefreshInterval.String(),
		"--lakom-config-path=" + lakomConfigDir + "/config.yaml",
		"--tls-cert-dir=" + webhookTLSCertDir,
		"--health-bind-address=:" + healthPort.String(),
		"--metrics-bind-address=:" + metricsPort.String(),
		"--port=" + serverPort.String(),
		"--use-only-image-pull-secrets=" + strconv.FormatBool(opts.useOnlyImagePullSecrets),
		"--insecure-allow-untrusted-images=" + strconv.FormatBool(opts.allowUntrustedImages),
		"--insecure-allow-insecure-registries=" + strconv.FormatBool(opts.allowInsecureRegistries),
	}

	if opts.authMode == authInjectedKubeconfig {
		args = append(args, "--kubeconfig="+gardenerutils.PathGenericKubeconfig)
	}
```

pkg/controller/lifecycle/lakom_resources.go
```diff
@@ -0,0 +1,851 @@
...
+							PreferredDuringSchedulingIgnoredDuringExecution: []corev1.WeightedPodAffinityTerm{{
+								Weight: 100,
+								PodAffinityTerm: corev1.PodAffinityTerm{
+									TopologyKey:   corev1.LabelHostname,
+									LabelSelector: &metav1.LabelSelector{MatchLabels: opts.podLabels},
+								},
+							}},
+						},
+					},
+					AutomountServiceAccountToken: ptr.To(automountServiceAccountToken),
```
vpnachev: ptr.To is useful to get pointers to right values, but you already have a variable and can use its address directly 


```suggestion
					AutomountServiceAccountToken: &automountServiceAccountToken,
```

pkg/controller/lifecycle/lakom_resources.go
```diff
@@ -0,0 +1,851 @@
...
+	authMode              lakomAuthMode
+	genericKubeconfigName string
+	accessSecretName      string
+	createNamespace       bool
+	createInClusterRBAC   bool
+	// tlsServerSecret, when set, is delivered into opts.namespace as part of the ManagedResource. This is
+	// used by the runtime-garden flavour: the secretsManager generates the cert in the extension namespace,
+	// but the pod runs in lakom-system, so gardener-resource-manager must copy the secret there. When nil,
+	// the server TLS secret is expected to already exist in opts.namespace (shoot/virtual flavours).
+	tlsServerSecret *corev1.Secret
```
vpnachev: It should be possible to configure the SecretManager to generate the secrets in the desired namespace so that they should not be copied later. Is there something that prevents this?

pkg/controller/lifecycle/lakom_resources.go
```diff
@@ -0,0 +1,851 @@
...
+		clientObjects = append(clientObjects, &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      opts.serverTLSSecretName,
+				Namespace: opts.namespace,
+				Labels:    getLabels(),
+			},
+			Type: opts.tlsServerSecret.Type,
+			Data: opts.tlsServerSecret.Data,
+		})
+	}
```
vpnachev: More clean approach would be to copy the `opts.tlsServerSecret` secret and modify the name and namespace (eventually append more labels) 

something like 


```go 
	if opts.tlsServerSecret != nil {
		secret := opts.tlsServerSecret.DeepCopy()
		secret.Name = opts.serverTLSSecretName
		secret.Namespace = opts.namespace
		secret.Labels = utils.MergeStringMaps(secret.GetLabels(), getLabels())

		clientObjects = append(clientObjects, secret)
	}
```

pkg/controller/lifecycle/lakom_resources.go
```diff
@@ -0,0 +1,851 @@
...
+			},
+		},
+	)
+
+	resources, err := registry.AddAllAndSerialize(clientObjects...)
+	if err != nil {
+		return nil, err
+	}
+
+	return resources, nil
```
vpnachev: Unless the error is wrapped with some more details, you can simply do 



```suggestion
	return registry.AddAllAndSerialize(clientObjects...)
```

pkg/controller/lifecycle/lakom_resources.go
```diff
@@ -0,0 +1,851 @@
...
+			},
+			RoleRef:  roleRef,
+			Subjects: subjects,
+		})
+	}
+
+	return roleBindings
+}
+
+type webhookVariant struct {
```
vpnachev:
```suggestion
type webhookOptions struct {
```

pkg/controller/lifecycle/lakom_resources.go
```diff
@@ -0,0 +1,851 @@
...
+func gardenRuntimeWebhookVariant() webhookVariant {
+	return webhookVariant{
+		registry:               managedresources.NewRegistry(kubernetes.SeedScheme, kubernetes.SeedCodec, kubernetes.SeedSerializer),
+		configName:             constants.GardenRuntimeWebhookConfigurationName,
+		useServiceClientConfig: true,
+		namespaceSelector: metav1.LabelSelector{MatchExpressions: []metav1.LabelSelectorRequirement{
+			{
+				Key:      corev1.LabelMetadataName,
+				Operator: metav1.LabelSelectorOpNotIn,
+				Values:   []string{constants.LakomSystemNamespace},
```
vpnachev: The `kube-system` namespace also has to be excluded.

pkg/controller/lifecycle/lakom_resources.go
```diff
@@ -0,0 +1,851 @@
...
+			},
+		}},
+		// Exclude Lakom runtime pods (part-of=shoot-lakom-service) to avoid Fail-policy self-deadlock
+		objectSelector: metav1.LabelSelector{MatchExpressions: []metav1.LabelSelectorRequirement{
+			{
+				Key:      "app.kubernetes.io/part-of",
+				Operator: metav1.LabelSelectorOpNotIn,
+				Values:   []string{constants.ExtensionType},
+			},
```
vpnachev: I think the object selector will be applied to any object subject to admission, i.e. not only pods but also extensions. This makes it possible an Extension resource labeled with `app.kubernetes.io/part-of=shoot-lakom-service` to skip the admission. 

Considering the lakom pods are deployed in a dedicated namespace, the namespace selector to exclude them should be sufficient.

pkg/controller/lifecycle/lakom_resources.go
```diff
@@ -0,0 +1,851 @@
...
+	}
+}
+
+// serviceBasedWebhookClientConfig builds a Service-based webhook client config
+func serviceBasedWebhookClientConfig(caBundle []byte, namespace, serviceName, path string) admissionregistrationv1.WebhookClientConfig {
+	return admissionregistrationv1.WebhookClientConfig{
+		Service: &admissionregistrationv1.ServiceReference{
+			Namespace: namespace,
+			Name:      serviceName,
+			Path:      ptr.To(path),
```
vpnachev:
```suggestion
			Path:     &path,
```

pkg/controller/lifecycle/lakom_resources.go
```diff
@@ -0,0 +1,851 @@
...
+		},
+		CABundle: caBundle,
+	}
+}
+
+// urlBasedWebhookClientConfig builds a URL-based webhook client config
+func urlBasedWebhookClientConfig(caBundle []byte, host, path string) admissionregistrationv1.WebhookClientConfig {
+	url := fmt.Sprintf("https://%s%s", host, path)
+	return admissionregistrationv1.WebhookClientConfig{
+		URL:      ptr.To(url),
```
vpnachev:
```suggestion
		URL:     &url,
```

pkg/secrets/config.go
```diff
@@ -45,3 +50,42 @@ func ConfigsFor(namespace string) []extensionssecretsmanager.SecretConfigWithOpt
 		},
 	}
 }
+
+func ConfigsForGarden() []extensionssecretsmanager.SecretConfigWithOptions {
```
vpnachev, line 55: Exported symbols should have a documentation.

pkg/controller/lifecycle/lakom_resources.go
```diff
@@ -0,0 +1,851 @@
+package lifecycle
```
vpnachev, line 5: Add license header.

pkg/controller/lifecycle/add.go
```diff
@@ -36,6 +37,10 @@ type AddOptions struct {
 	ServiceConfig config.Configuration
 	// IgnoreOperationAnnotation specifies whether to ignore the operation annotation or not.
 	IgnoreOperationAnnotation bool
+	// ExtensionClasses contains the extension classes the controller should reconcile.
+	// Only a single type of extension class is supported at the moment.
+	// Depending on the extension class, the controller will target shoot control plane or garden namespaces.
```
vpnachev: I do not understand this, this PR goal is to add support for the Garden class next to the already existing Shoot class, so it should be supporting single type only, no?

`shoot control plane` and `garden namespace` are not very precise considering what the extension is doing, in the `Shoot` class case - the target is the workload running in the shoot (the lakom service is hosted in the shoot control plane namespace). In the Garden class, the target is the workload and configuration resource in the both garden clusters - runtime and virtual, in any case not restricted to the `garden` namespace only.

pkg/controller/lifecycle/lakom_resources.go
```diff
@@ -0,0 +1,851 @@
...
+			Operations: []admissionregistrationv1.OperationType{admissionregistrationv1.Create, admissionregistrationv1.Update},
+			Rule: admissionregistrationv1.Rule{
+				APIGroups:   []string{"operator.gardener.cloud"},
+				APIVersions: []string{"v1alpha1"},
+				Resources:   []string{"extensions"},
+			},
+		},
+	}
+
+	gardenWebhookVirtualGardenRules []admissionregistrationv1.RuleWithOperations = []admissionregistrationv1.RuleWithOperations{
```
vpnachev:
```suggestion
	virtualGardenWebhookRules []admissionregistrationv1.RuleWithOperations = []admissionregistrationv1.RuleWithOperations{
```

to avoid duplication of the term `garden`.

pkg/controller/lifecycle/lakom_resources.go
```diff
@@ -0,0 +1,851 @@
...
+	shootWebhookRules []admissionregistrationv1.RuleWithOperations = []admissionregistrationv1.RuleWithOperations{{
+		Operations: []admissionregistrationv1.OperationType{admissionregistrationv1.Create, admissionregistrationv1.Update},
+		Rule: admissionregistrationv1.Rule{
+			APIGroups:   []string{""},
+			APIVersions: []string{"v1"},
+			Resources:   []string{"pods", "pods/ephemeralcontainers"},
+		},
+	}}
+
+	gardenWebhookRuntimeRules []admissionregistrationv1.RuleWithOperations = []admissionregistrationv1.RuleWithOperations{
```
vpnachev:
```suggestion
	runtimeGardenWebhookRules []admissionregistrationv1.RuleWithOperations = []admissionregistrationv1.RuleWithOperations{
```

to align it with `virtualGardenWebhookRules`

pkg/controller/lifecycle/lakom_resources.go
```diff
@@ -0,0 +1,851 @@
...
+				APIVersions: []string{"v1alpha1"},
+				Resources:   []string{"gardenlets"},
+			},
+		},
+	}
+)
+
+func getWebhookResources(
+	webhookVariant webhookVariant,
+	webhookCaBundle []byte,
```
vpnachev: What do you think about making the ca bundle part of the `webhookVariant` (`webhookOptions`)?

pkg/controller/lifecycle/lakom_resources.go
```diff
@@ -0,0 +1,851 @@
...
+	webhookRRules []admissionregistrationv1.RuleWithOperations,
+	serviceName,
+	extensionNamespace string,
+) (map[string][]byte, error) {
+	clientConfigFor := func(path string) admissionregistrationv1.WebhookClientConfig {
+		if webhookVariant.useServiceClientConfig {
+			return serviceBasedWebhookClientConfig(webhookCaBundle, extensionNamespace, serviceName, path)
+		}
+		return urlBasedWebhookClientConfig(webhookCaBundle, fmt.Sprintf("%s.%s", serviceName, extensionNamespace), path)
```
vpnachev: Both functions receive the same parameter values (module some formatting). This could become 
```suggestion
		return getWebhookClientConfig(webhookVariant.useServiceClientConfig, webhookCaBundle, serviceName, extensionNamespace, path)
```


where `serviceBasedWebhookClientConfig` and `urlBasedWebhookClientConfig` are replaced by `getWebhookClientConfig`


```go
// getClientConfig builds a webhook client config
func getWebhookClientConfig(useServiceClientConfig bool, caBundle []byte, namespace, serviceName, path string) admissionregistrationv1.WebhookClientConfig {
	if useServiceClientConfig {
		return admissionregistrationv1.WebhookClientConfig{
			Service: &admissionregistrationv1.ServiceReference{
				Namespace: namespace,
				Name:      serviceName,
				Path:      &path,
			},
			CABundle: caBundle,
		}
	}

	url := fmt.Sprintf("https://%s.%s%s", serviceName, namespace, path)
	return admissionregistrationv1.WebhookClientConfig{
		URL:      &url,
		CABundle: caBundle,
	}
}
```

pkg/controller/lifecycle/lakom_resources.go
```diff
@@ -0,0 +1,851 @@
...
+	}
+
+	var (
+		matchPolicy       = admissionregistrationv1.Equivalent
+		sideEffectClass   = admissionregistrationv1.SideEffectClassNone
+		failurePolicy     = admissionregistrationv1.Fail
+		timeOutSeconds    = ptr.To[int32](25)
+		namespaceSelector = webhookVariant.namespaceSelector
+		objectSelector    = webhookVariant.objectSelector
+		clientObjects     = []client.Object{
```
vpnachev:
```suggestion
		resources     = []client.Object{
```

`clientObjects` does not sound right when the function purpose is to construct the resources.

pkg/controller/lifecycle/lakom_resources.go
```diff
@@ -0,0 +1,851 @@
...
+
+	}
+
+	shootResources, err := webhookVariant.registry.AddAllAndSerialize(clientObjects...)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return shootResources, nil
```
vpnachev:
```suggestion
	return webhookVariant.registry.AddAllAndSerialize(resources...)
```

pkg/controller/lifecycle/lakom_resources.go
```diff
@@ -0,0 +1,851 @@
...
+type lakomAuthMode int
+
+const (
+	// authInjectedKubeconfig injects a generic kubeconfig (shoot-access) into the deployment and points
+	// lakom at it via --kubeconfig.
+	authInjectedKubeconfig lakomAuthMode = iota
+	// authInClusterToken relies on the pod's mounted ServiceAccount token (in-cluster auth) and omits
+	// the --kubeconfig flag.
+	authInClusterToken
+)
```
vpnachev: This can be replaced by a single boolean parameter, is there any benefit of using enum today?

pkg/controller/lifecycle/lakom_resources.go
```diff
@@ -0,0 +1,851 @@
...
+				},
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: utils.MergeStringMaps(opts.podLabels, map[string]string{
+						v1beta1constants.LabelNetworkPolicyToDNS:                                                                    v1beta1constants.LabelNetworkPolicyAllowed,
+						v1beta1constants.LabelNetworkPolicyToPublicNetworks:                                                         v1beta1constants.LabelNetworkPolicyAllowed,
+						v1beta1constants.LabelNetworkPolicyToPrivateNetworks:                                                        v1beta1constants.LabelNetworkPolicyAllowed,
+						v1beta1constants.LabelNetworkPolicyToBlockedCIDRs:                                                           v1beta1constants.LabelNetworkPolicyAllowed,
+						gardenerutils.NetworkPolicyLabel(v1beta1constants.DeploymentNameKubeAPIServer, kubeapiserverconstants.Port): v1beta1constants.LabelNetworkPolicyAllowed,
```
vpnachev, line 245: We have to double check if these network labels are not allowing too much network access in the garden class scenario.

pkg/controller/lifecycle/lakom_resources.go
```diff
@@ -0,0 +1,851 @@
...
+			Labels:    getLabels(),
+		},
+		Spec: policyv1.PodDisruptionBudgetSpec{
+			MaxUnavailable:             ptr.To(intstr.FromInt32(1)),
+			Selector:                   &metav1.LabelSelector{MatchLabels: opts.podLabels},
+			UnhealthyPodEvictionPolicy: ptr.To(policyv1.AlwaysAllow),
+		},
+	}
+
+	clientObjects := []client.Object{}
```
vpnachev:
```suggestion
	resources := []client.Object{}
```

pkg/controller/lifecycle/actuator.go
```diff
@@ -117,122 +120,256 @@ func (a *actuator) Reconcile(ctx context.Context, logger logr.Logger, ex *extens
...
+
+	timeoutSeedCtx, cancelSeedCtx := context.WithTimeout(ctx, 2*time.Minute)
 	defer cancelSeedCtx()
-	if err := managedresources.WaitUntilHealthy(timeoutSeedCtx, a.client, namespace, constants.ManagedResourceNamesSeed); err != nil {
+	if err := managedresources.WaitUntilHealthy(timeoutSeedCtx, a.client, clusterCtx.namespace, constants.ManagedResourceNamesGardenRuntime); err != nil {
 		return err
 	}
 
-	return secretsManager.Cleanup(ctx)
+	if err := managedresources.WaitUntilHealthy(timeoutSeedCtx, a.client, clusterCtx.namespace, constants.ManagedResourceNamesGardenVirtual); err != nil {
```
vpnachev: Each `WaitUntilHealthy`  call should use own context with timeout, otherwise they share the same timeout, which means it is expected all 3 managed resources are expected to get healthy in 2 minutes in total instead of each having 2 minutes.

pkg/controller/lifecycle/actuator.go
```diff
@@ -117,122 +120,256 @@ func (a *actuator) Reconcile(ctx context.Context, logger logr.Logger, ex *extens
...
+
+	if err := managedresources.WaitUntilHealthy(timeoutSeedCtx, a.client, clusterCtx.namespace, constants.ManagedResourceNamesGardenRuntimeWebhook); err != nil {
+		return err
+	}
+	// Wait for the ManagedResources.
+	// Cleanup secrets manager after they are healthy.
+	return clusterCtx.secretsManager.Cleanup(ctx)
+}
+
+func (a *actuator) prepareAssets(
```
vpnachev: This function seems to be doing a lot of this like generating secrets, modifying the clusterContext. For me `prepareAssets` is too generic and it is hard to get an idea what to expect from the function, a better name and/or doc string should be implemented.

pkg/controller/lifecycle/actuator.go
```diff
@@ -308,516 +521,179 @@ func getLabels() map[string]string {
...
+		return nil, fmt.Errorf("could not get runtime cluster server version: %w", err)
 	}
 
-	refSecret := &corev1.Secret{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      ref.ResourceRef.Name,
-			Namespace: namespace,
-		},
+	if _, err := semver.NewVersion(versionInfo.GitVersion); err != nil {
+		return nil, fmt.Errorf("failed to parse runtime cluster Kubernetes version %q: %w", versionInfo.GitVersion, err)
```
vpnachev: Considering the parsed semver versioin is not used, I would skip parsing it here.

pkg/controller/healthcheck/add.go
```diff
@@ -32,6 +32,23 @@ var (
...
+		managedResourceNames = []string{
+			constants.ManagedResourceNamesGardenRuntime,
+			constants.ManagedResourceNamesGardenVirtual,
+		}
+	}
+
+	healthChecks := make([]healthcheck.ConditionTypeToHealthCheck, 0, len(managedResourceNames))
+	for _, managedResourceName := range managedResourceNames {
+		healthChecks = append(healthChecks, healthcheck.ConditionTypeToHealthCheck{
+			ConditionType: string(gardencorev1beta1.ShootControlPlaneHealthy),
```
vpnachev: Does shoot control plane healthy work well for Garden? There is another condition that probably fits better the garden use case - `github.com/gardener/gardener/pkg/apis/operator/v1alpha1.RuntimeComponentsHealthy`

charts/gardener-extension-shoot-lakom-service/templates/clusterparsers-logging.yaml
```diff
@@ -6,7 +6,7 @@
 apiVersion: fluentbit.fluent.io/v1alpha2
 kind: ClusterParser
 metadata:
-  name: extension-shoot-lakom-service-lakom-parser
+  name: extension-shoot-lakom-service-lakom-parser-{{ include "name" . }}
```
vpnachev:
```suggestion
  name: extension-{{ include "name" . }}-lakom-parser
```

This should keep the name shorter and avoid duplications. Otherwise, the resource is named like `extension-shoot-lakom-service-lakom-parser-shoot-lakom-service(-runtime)`

charts/gardener-extension-shoot-lakom-service/templates/clusterfilters-logging.yaml
vpnachev: I guess this resource also needs its name to be changed for the garden class extension. 

```suggestion
  name: extension-{{ include "name" . }}-lakom
```

charts/gardener-extension-shoot-lakom-service/templates/deployment.yaml
```diff
@@ -84,6 +84,14 @@ spec:
         - --health-bind-address=:{{ .Values.controllers.healthPort }}
         - --metrics-bind-address=:{{ .Values.controllers.metricsPort }}
         - --seed-topology-aware-routing-enabled={{ .Values.gardener.seed.spec.settings.topologyAwareRouting.enabled }}
+        {{- if .Values.gardener.runtimeCluster.enabled }}
+        # Runtime deployments serve Garden-class Extension resources; 
+        #keep the legacy seed bootstrap controller only in seed deployments to avoid managing the same kube-system resources from two namespaces.
```
vpnachev:
```suggestion
```

I would remove this comment, disabling the controller is self-explanatory. In the garden class case it is not expected the controller to be running in a seed cluster, respectively the controller should be disabled.

charts/gardener-extension-shoot-lakom-service/templates/rbac.yaml
vpnachev, line 12:
```suggestion
rules:
{{- if not .Values.gardener.runtimeCluster.enabled }}
```

charts/gardener-extension-shoot-lakom-service/templates/rbac.yaml
```diff
@@ -18,6 +18,16 @@ rules:
   - get
   - list
   - watch
+{{- if .Values.gardener.runtimeCluster.enabled }}
```
vpnachev:
```suggestion
{{- else }}
```

charts/gardener-extension-shoot-lakom-service/templates/rbac.yaml
```diff
@@ -35,6 +45,8 @@ rules:
   resources:
   - namespaces
   verbs:
+  - create
+  - delete
```
vpnachev, line 51:
```suggestion
  {{- if .Values.gardener.runtimeCluster.enabled }}
  - create
  - delete
  {{- end }}
```

charts/gardener-extension-shoot-lakom-service/templates/rbac.yaml
```diff
@@ -67,6 +79,32 @@ rules:
...
+  name: extensions.gardener.cloud:{{ include "name" . }}
+subjects:
+- kind: ServiceAccount
+  name: gardener-extension-{{ .Values.serviceName }}
+  namespace: {{ .Release.Namespace }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: gardener-extension-{{ .Values.serviceName }}-leader-election
```
vpnachev: In the future the controller might need access to other resource in the namespace, therefore let's not put the `leader-election` in its name


```suggestion
  name: extensions.gardener.cloud:{{ include "name" . }}
```

charts/gardener-extension-shoot-lakom-service/templates/rbac.yaml
```diff
@@ -88,16 +126,17 @@ rules:
   - update
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRoleBinding
+kind: RoleBinding
 metadata:
-  name: extensions.gardener.cloud:{{ .Values.serviceName }}
+  name: gardener-extension-{{ .Values.serviceName }}-leader-election
```
vpnachev:
```suggestion
  name: extensions.gardener.cloud:{{ include "name" . }}
```

charts/gardener-extension-shoot-lakom-service/templates/rbac.yaml
```diff
@@ -88,16 +126,17 @@ rules:
...
+  namespace: {{ .Release.Namespace }}
   labels:
     app.kubernetes.io/name: gardener-extension-{{ .Values.serviceName }}
     app.kubernetes.io/instance: {{ .Release.Name }}
 roleRef:
   apiGroup: rbac.authorization.k8s.io
-  kind: ClusterRole
-  name: extensions.gardener.cloud:{{ .Values.serviceName }}
+  kind: Role
+  name: gardener-extension-{{ .Values.serviceName }}-leader-election
```
vpnachev:
```suggestion
  name: extensions.gardener.cloud:{{ include "name" . }}
```

pkg/controller/lifecycle/actuator.go
```diff
@@ -99,16 +87,30 @@ func getLakomReplicas(hibernated bool) *int32 {
 	return ptr.To[int32](3)
 }
 
-// Reconcile the Extension resource.
+// Reconcile the Extension resource
```
vpnachev:
```suggestion
// Reconcile the Extension resource.
```

Please, keep the punctuation, ref: https://go.dev/wiki/CodeReviewComments#comment-sentences

pkg/controller/lifecycle/actuator.go
```diff
@@ -308,516 +450,256 @@ func getLabels() map[string]string {
...
 
-	clientKeys, ok := refSecret.Data["keys"]
-	if !ok {
-		return nil, fmt.Errorf("secret %s/%s is missing data key 'keys'", refSecret.Namespace, refSecret.Name)
+	image, err := imagevector.ImageVector().FindImage(constants.ImageName)
+	if err != nil {
+		return fmt.Errorf("failed to find image version for %s: %v", constants.ImageName, err)
+	}
+	if image.Tag == nil {
+		image.Tag = ptr.To[string](version.Get().GitVersion)
```
vpnachev:
```suggestion
		image.Tag = ptr.To(version.Get().GitVersion)
```

in this case it is not necessary to explicitly specify the type.

pkg/controller/lifecycle/actuator.go
```diff
@@ -308,516 +450,256 @@ func getLabels() map[string]string {
...
+			clusterCtx.namespace,
+		)
+		if err != nil {
+			return fmt.Errorf("failed to get the additional keys: %w", err)
+		}
+	}
+	lakomPublicKeys := make([]byte, 0, len(gardenerPublicKeys)+len(clientPublicKeys))
+	lakomPublicKeys = append(lakomPublicKeys, gardenerPublicKeys...)
+	lakomPublicKeys = append(lakomPublicKeys, clientPublicKeys...)
+	clusterCtx.lakomPublicKeysConfig = lakomPublicKeys
```
vpnachev: This can be simplified to 


```suggestion
	clusterCtx.lakomPublicKeysConfig = append(gardenerPublicKeys, clientPublicKeys...)
```

pkg/controller/lifecycle/lakom_resources.go
```diff
@@ -0,0 +1,779 @@
+// SPDX-FileCopyrightText: 2022 SAP SE or an SAP affiliate company and Gardener contributors
```
vpnachev:
```suggestion
// SPDX-FileCopyrightText: SAP SE or an SAP affiliate company and Gardener contributors
```

There is a not very well documented ongoing initiative to remove the year from the license header

pkg/controller/lifecycle/lakom_resources.go
```diff
@@ -0,0 +1,779 @@
...
+				APIVersions: []string{"v1alpha1"},
+				Resources:   []string{"gardenlets"},
+			},
+		},
+	}
+)
+
+func getWebhookResources(
+	webhookOptions webhookOptions,
+	webhookRRules []admissionregistrationv1.RuleWithOperations,
```
vpnachev:
```suggestion
	webhookRules []admissionregistrationv1.RuleWithOperations,
```

I guess this is a typo?

pkg/controller/lifecycle/lakom_resources.go
```diff
@@ -0,0 +1,779 @@
...
+				Rules: []rbacv1.PolicyRule{
+					{
+						APIGroups: []string{""},
+						Resources: []string{"secrets"},
+						Verbs:     []string{"get"},
+					},
+				},
+			})
+		resources = append(resources,
+			getRoleBindings(webhookOptions.scope, webhookOptions.resourceReaderSvcAccName, webhookOptions.dashboardEnabled)...)
```
vpnachev:
```suggestion
			getRoleBindings(webhookOptions.scope, webhookOptions.resourceReaderSvcAccName, webhookOptions.dashboardEnabled)...,
		)
```

pkg/controller/lifecycle/lakom_resources.go
```diff
@@ -0,0 +1,779 @@
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      opts.serviceName,
+				Namespace: opts.namespace,
+				Labels:    getLabels(),
+			},
+			AutomountServiceAccountToken: &opts.useInClusterAuth,
+		},
+	)
+	if opts.createInClusterRBAC {
+		resources = append(resources,
```
vpnachev: In which scenario are these Role and RoleBinding used? I am wondering how is this aligned with the `getRoleBindings` func.

cmd/gardener-extension-shoot-lakom-service/app/app.go
```diff
@@ -109,6 +112,13 @@ func (o *Options) run(ctx context.Context) error {
 			},
 		},
 	}
+	mgrOpts.Cache.ByObject = map[client.Object]cache.ByObject{
+		&coordinationv1.Lease{}: {
+			Namespaces: map[string]cache.Config{
+				mgrOpts.LeaderElectionNamespace: {},
+			},
+		},
+	}
```
vpnachev, line 124: Can you add a comment why is this needed?

pkg/controller/lifecycle/actuator.go
h-mavrodiev: This is how it was written initially, but it was picked up by the linter, which caused the "verify" pipeline to fail.

charts/gardener-extension-shoot-lakom-service/templates/clusterfilters-logging.yaml
```diff
@@ -8,11 +8,11 @@ kind: ClusterFilter
...
   labels:
     fluentbit.gardener/type: "seed"
-  name: extension-shoot-lakom-service-lakom
+  name: extension-{{ include "name" . }}-lakom
 spec:
   filters:
   - parser:
       keyName: log
-      parser: extension-shoot-lakom-service-lakom-parser
+      parser: extension-shoot-lakom-service-lakom-parser-{{ include "name" . }}
```
vpnachev, line 16:
```suggestion
      parser: extension-{{ include "name" . }}-lakom-parser
```

pkg/controller/lifecycle/actuator.go
```diff
@@ -117,126 +121,195 @@ func (a *actuator) Reconcile(ctx context.Context, logger logr.Logger, ex *extens
...
 	}
 
-	var clientPublicKeys []byte
-	if lakomProviderConfig.TrustedKeysResourceName != nil {
-		var err error
-		clientPublicKeys, err = getClientKeys(ctx, a.client, cluster.Shoot.Spec.Resources, *lakomProviderConfig.TrustedKeysResourceName, namespace)
-		if err != nil {
-			return fmt.Errorf("failed to get the additional keys: %w", err)
-		}
+	gardenAccessSecret := gardenerutils.NewGardenAccessSecret(
```
vpnachev, line 190: I think the service account namespace will have to be explicitly set, I am testing locally right now and the namespace is `seed-local` which is used to store copies of secrets (and other resources) related to the `local` seed, but we are looking for access to the garden cluster, hence it should be created in the `garden` namespace, i.e. 

```go
gardenAccessSecret.WithServiceAccountNamespace(clusterCtx.namespace)
```

pkg/controller/lifecycle/lakom_resources.go
```diff
@@ -0,0 +1,763 @@
...
+	roleRef := rbacv1.RoleRef{
+		APIGroup: "rbac.authorization.k8s.io",
+		Kind:     "ClusterRole",
+		Name:     constants.LakomResourceReader,
+	}
+	subjects := []rbacv1.Subject{
+		{
+			Kind:      rbacv1.ServiceAccountKind,
+			Name:      shootAccessServiceAccountName,
+			Namespace: metav1.NamespaceSystem,
```
vpnachev, line 606: The namespace should be configurable, for the virtual cluster the service account should be in the `garden` namespace, for example.

pkg/controller/lifecycle/lakom_resources.go
```diff
@@ -0,0 +1,763 @@
...
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	vpaautoscalingv1 "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	shootWebhookRules []admissionregistrationv1.RuleWithOperations = []admissionregistrationv1.RuleWithOperations{{
```
vpnachev, line 40: You can omit the type, it is derived from the value
```suggestion
	shootWebhookRules = []admissionregistrationv1.RuleWithOperations{{
```

pkg/controller/lifecycle/actuator_test.go
```diff
@@ -55,6 +60,30 @@ var _ = Describe("Actuator", func() {
 		Expect(appPartOf).To(Equal("shoot-lakom-service"))
 	})
 
+	DescribeTable("Should get the garden pod labels with a variant-specific instance label",
+		func(virtualGarden bool) {
+			labels := getGardenPodLabels(virtualGarden)
+			expectedInstance := constants.GardenRuntimeExtensionServiceName
```
vpnachev, line 66: In tests, it is a good practice to use the literal instead of the const, in this way the test should fail if the const value is changed


```suggestion
			expectedInstance := "extension-shoot-lakom-service-garden-runtime"
```

pkg/controller/lifecycle/actuator_test.go
```diff
@@ -428,8 +548,256 @@ var _ = Describe("Actuator", func() {
...
+			lakomConfigConfigMapName = "extension-shoot-lakom-service-garden-virtual-lakom-config-5ccba116"
+		)
+
+		var (
+			replicas    int32
+			lakomConfig string
+		)
+
+		BeforeEach(func() {
+			replicas = int32(3)
```
vpnachev, line 568: The variable has a type assigned, there is no need to explicit here
```suggestion
			replicas = 3
```

pkg/controller/lifecycle/actuator_test.go
```diff
@@ -428,8 +548,256 @@ var _ = Describe("Actuator", func() {
...
+					useOnlyImagePullSecrets,
+					allowUntrustedImages,
+					allowInsecureRegistries,
+				)
+				Expect(err).ToNot(HaveOccurred())
+				resources := serializeSeedObjects(objects)
+				Expect(resources).To(HaveKey("data.yaml.br"))
+				compressedData := resources["data.yaml.br"]
+				data, err := test.BrotliDecompression(compressedData)
+				Expect(err).NotTo(HaveOccurred())
```
vpnachev, line 614: Not an issue per se, but be consistent when using `NotTo` or `ToNot`, both do the same, but for consistency it would be nice only one to be used.

pkg/controller/lifecycle/actuator_test.go
```diff
@@ -848,3 +1420,556 @@ spec:
...
+  unhealthyPodEvictionPolicy: AlwaysAllow
+status:
+  currentHealthy: 0
+  desiredHealthy: 0
+  disruptionsAllowed: 0
+  expectedPods: 0
+`
+}
+
+func expectedGardenVirtualService(namespace string) string {
```
vpnachev, line 1598: Looks like the namespace is constant, in this case I think it does not need to be a parameter.

pkg/controller/lifecycle/actuator.go
```diff
@@ -117,126 +121,193 @@ func (a *actuator) Reconcile(ctx context.Context, logger logr.Logger, ex *extens
...
+		clusterCtx.caBundle,
+	)
+	shootObjects, err := getWebhookObjects(shootWebhookOptions, shootWebhookRules, constants.ExtensionServiceName, clusterCtx.namespace)
 	if err != nil {
 		return err
 	}
-
-	generatedSecrets, err := extensionssecretsmanager.GenerateAllSecrets(ctx, secretsManager, configs)
+	shootObjects = append(shootObjects,
+		getTargetClusterRBACObjects(*clusterCtx.providerConfig.Scope, lakomShootAccessSecret.ServiceAccountName, clusterCtx.namespace, clusterCtx.dashboardEnabled)...,
```
vpnachev:
```suggestion
		getTargetClusterRBACObjects(*clusterCtx.providerConfig.Scope, lakomShootAccessSecret.ServiceAccountName, metav1.NamespaceSystem, clusterCtx.dashboardEnabled)...,
```

vpnachev (commented):
Just one issue with the namespace for the in-shoot resources, rest looks good to me.

Issue #123, opened 2025-01-07
E2E Tests


rrhubenov:
**What would you like to be added**:
Add a suite of e2e tests.
**Why is this needed**:
More reliable error catching and feature development using automated e2e tests.

PR #336, opened 2026-08-11
Add e2e test make command and script placeholders

h-mavrodiev:

/area testing
/kind enhancement

Adds an e2e make command and script placeholders to help validate end-to-end tests currently under development.

Related to https://github.com/gardener/gardener-extension-shoot-lakom-service/issues/123

NONE
**Release note**:

```other operator
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -154,6 +154,10 @@ test-cov:
 test-clean:
 	@bash $(GARDENER_HACK_DIR)/test-cover-clean.sh
 
+.PHONY: ci-e2e-kind
+ci-e2e-kind: $(KIND) $(YQ)
+	./hack/ci-e2e-kind.sh
+
 .PHONY: verify
 verify: check format test sast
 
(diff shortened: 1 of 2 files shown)


hack/ci-e2e-kind.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
vpnachev:
```suggestion
#!/usr/bin/env bash
```

Some environments have the `bash` available on different path, it turned out that more reliable is `env` which we so far have not seen to be unavailable under `/usr/bin`

Makefile
```diff
@@ -154,6 +154,10 @@ test-cov:
 test-clean:
 	@bash $(GARDENER_HACK_DIR)/test-cover-clean.sh
 
+.PHONY: ci-e2e-kind
+ci-e2e-kind: $(KIND) $(YQ)
```
vpnachev: `kind` and `yq` are not needed as of now, they can be added later.

```suggestion
ci-e2e-kind:
```

